Issue #2, opened 2019-08-20
Abstract generic base class and model


sisp:
I'm trying to implement a generic base class with a generic model prop but some shared logic among all subclasses (here the `error` getter):

```ts
import { computed } from 'mobx';
import { model, Model, prop } from 'mobx-keystone';

abstract class A<P> extends Model({
  value: prop<P>(), // Base class expressions cannot reference class type parameters. ts(2562)
}) {
  public abstract validate(value: P): string | undefined;

  @computed
  public get error(): string | undefined {
    return this.validate(this.value);
  }
}

@model('B')
class B extends A<string> {
  public validate(value: string): string | undefined {
    return value.length < 3 ? 'too short' : undefined;
  }
}

```

How would you implement this instead? Using the new `ExtendedModel` factory doesn't seem to help because I wouldn't be able to declare the generic prop and abstract `validate` method in `A`, so the `error` getter couldn't be implemented already in `A`.

Any help is much appreciated. :-)

sisp:
Perhaps the generic model `prop` is not necessary (besides the fact that this cannot work anyway) because declaring an abstract model prop is only a typing issue, so the abstract prop could be declared as a regular class property. Something like this could be a potential solution:

```ts
import { computed } from 'mobx';
import { ExtendedModel, model, Model, prop } from 'mobx-keystone';

abstract class A<P> extends Model({}) {
  public abstract value: P;

  public abstract validate(value: P): string | undefined;

  @computed
  public get error(): string | undefined {
    return this.validate(this.value);
  }
}

type BValue = string;

abstract class _B extends A<BValue> {}

// Argument of type 'typeof B' is not assignable to parameter of type 'ModelClass<BaseModel<any, any>>'.
//   Type 'B' is missing the following properties from type 'BaseModel<any, any>': $, typeCheck, [typeSymbol], [modelTypeKey] ts(2345)
@model('B')
class B extends ExtendedModel(
  // Argument of type 'typeof _B' is not assignable to parameter of type 'ModelClass<_B>'.
  //   Cannot assign an abstract constructor type to a non-abstract constructor type. ts(2345)
  _B,
  {
    value: prop<BValue>(),
  }
) {
  public validate(value: BValue): string | undefined {
    return value.length < 3 ? 'too short' : undefined;
  }
}
```

[...]

PR #9, opened 2019-08-27
Subclassing with abstract class

sisp:
Regarding extending an abstract base class with `ExtendedModel` (https://github.com/xaviergonz/mobx-keystone/issues/2#issuecomment-523160527):

How about using [this suggestion](https://stackoverflow.com/a/38642922) for typing an abstract class:

```ts
type Constructor<T> = Function & { prototype: T }
```

The problem with using `ModelClass<T>` for typing abstract classes is that abstract classes don't have `new (...): ...` because they cannot be instantiated.

### DIFF
--- a/packages/lib/src/model/Model.ts
+++ b/packages/lib/src/model/Model.ts
@@ -50,10 +50,10 @@ export interface _Model<TProps extends ModelProps> {
  * @returns
  */
 export function ExtendedModel<TProps extends ModelProps, TBaseModel extends AnyModel>(
-  baseModel: ModelClass<TBaseModel>,
+  baseModel: Function & { prototype: TBaseModel },
   modelProps: TProps
 ): _ExtendedModel<TBaseModel, TProps> {
-  return internalModel<TProps, TBaseModel>(modelProps, baseModel)
+  return internalModel<TProps, TBaseModel>(modelProps, baseModel as ModelClass<TBaseModel>)
 }
 
 /**
--- a/packages/lib/test/model/subclassing.test.ts
+++ b/packages/lib/test/model/subclassing.test.ts
@@ -301,3 +301,34 @@ test("abstract-ish model classes", () => {
   expect(b.error).toBe("too short")
   expect(b instanceof StringA).toBe(true)
 })
+
+test("abstract model classes", () => {
+  abstract class A<P> extends Model({}) {
+    public abstract value: P
+
+    public abstract validate(_value: P): string | undefined
+
+    @computed
+    public get error(): string | undefined {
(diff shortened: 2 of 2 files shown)


xaviergonz:
I think I tried that, but after that change if you check the type of the instance props you'll see they became any IIRC

sisp:
I checked the types in the new test I added in VS Code (e.g. `b.value`, `b.validate`) and they are what they're supposed to be, not `any`. Also the class constructor's `initialData` object appears to be typed correctly. Could you please double-check?

xaviergonz:
Sorry, I meant when using a non abstract class as base with that change the value and error for example become any (the test above it), but that gave me an idea, let me try something...

sisp:
I still don't seem to be able to reproduce your problem.

Let's say the test looked like this (with a non-abstract base class passed to `ExtendedModel`):

```ts
test("abstract model classes", () => {
  class A extends Model({
    value: prop<string>(),
  }) {
    public validate?(_value: string): string | undefined

    @computed
    public get error(): string | undefined {
      return this.validate!(this.value)
    }
  }

  @model("B")
  class B extends ExtendedModel(A, {
    value: prop<string>(),
  }) {
    public validate(value: string): string | undefined {
      return value.length < 3 ? "too short" : undefined
    }
  }

  const b = new B({ value: "hi" })
  expect(b.value).toBe("hi")
  expect(b.validate("ho")).toBe("too short")
  expect(b.validate("long")).toBe(undefined)
  expect(b.error).toBe("too short")
  expect(b instanceof A).toBe(true)
})
```

Then the type of `b.value` is still `string`. I can't find `any` anywhere.

xaviergonz:
I created this PR https://github.com/xaviergonz/mobx-keystone/pull/12 based on yours but with some extra cases
The failing case was the "abstract model classes with factory", where the prop "value" was taking the value of any since by default the type of the prototype of a generic class C<T> takes the type of C<any>, so the only way I found to solve that case was to typecast the result of the factory function like this

```
    // we need this weird trick to make value get the right type in this case
    return A as typeof A & {
      new (): A
    }
```

since new() actually contains the proper generic in the type

besides for that particular case (a factory that returns a generic abstract class with a prop that is a generic in itself) everything else seems fine :)

sisp:
Looks great! I think the type hack for the return type of the factory is not too bad. I really appreciate your efforts to support abstract base classes as this makes my models much cleaner. :-)

Perhaps it makes sense to document the different options similar to the tests you implemented, so people can choose what works best for them and become aware of the type hack, too.

xaviergonz:
closed in favor for the other pr

PR #12, opened 2019-08-28
improved support for abstract base classes

xaviergonz:


sisp:
Looks great! I really appreciate your efforts to support abstract base classes as this makes my models much cleaner. :-)

PR #13, opened 2019-08-30
Fix ESM bundle filename and remove UMD bundle reference

sisp:
The ESM bundle filename in `packages/lib/package.json` does not match the actually built filename and there is a reference to a UMD bundle which currently is not built. Would you like to build a UMD bundle, too? This needs to be explicitly enabled since [TSDX v0.7.0](https://github.com/palmerhq/tsdx/releases/tag/v0.7.0).


sisp:
Hmm, I screwed something up. I'll create a new PR. Sorry ...

PR #16, opened 2019-08-30
getParentToChildPath, getChildrenObjects, onChildAttachedTo

xaviergonz:


packages/lib/src/parent/onChildAttachedTo.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const currentChildren = opts.runForCurrentChildren ? new Set<object>() : getCurrentChildren()
+
+  const disposer = reaction(
+    () => getCurrentChildren(),
+    newChildren => {
+      const disposersToRun: object[] = []
+
+      // find dead
+      currentChildren.forEach((n: object) => {
```
terrysahaidak: Hey @xaviergonz, have you ever tried to measure performance in such a case between classic for-loop and forEach? I'm just wonder :)

xaviergonz: not for sets, that's for sure. I'd need to check it against the TS transpilation though, since TS downlevel iteration does some heavy stuff, thanks for the tip!

xaviergonz: Benchmark
foreach: 8,141 ops/sec 
for of (ES6):  49,881 ops/sec
for of (ES5 with downlevel iteration): 13,534 ops/sec

So guess I'll change it to for of and if there's ever an ES6 version of the library it should get even faster, thanks!

xaviergonz: Found this is even faster, 50k ops
    const iter = set.values()
    let cur
    while (!(cur = iter.next()).done) {
      const n = cur.value
    }

terrysahaidak: The last one is really weird looking one :D

xaviergonz: Whatever works fast :D

xaviergonz:
will be released as 0.14

PR #17, opened 2019-08-31
try to fix inheritance for es6

### LINKED ISSUE #15: create-react-app: TypeError: Class constructor [...] cannot be invoked without 'new'
When using `ExtendedModel` in a `create-react-app`-based app, I'm getting the following error:

> TypeError: Class constructor [...] cannot be invoked without 'new'

This is a minimal reproduction: https://github.com/sisp/mobx-keystone-cra-issue

The error does not occur when I run the example with `ts-node`, so I think the problem originates from CRA's usage of Babel + Typescript build chain.

xaviergonz:
Fixes #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 0.13.1
+
+- Fixed an issue with ExtendsModel when user library was compiled using ES6 classes.
+
 ## 0.13.0
 
 - Improved support for abstract base models.
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx-keystone",
-  "version": "0.13.0",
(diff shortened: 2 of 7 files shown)


xaviergonz:
@sisp care to give this build a try and see if it fixes your CRA issue?

sisp:
When I use the ESM bundle built from this branch with my [CRA example](https://github.com/sisp/mobx-keystone-cra-issue), it works (for both `"target": "es5"` and `"target": "es6"` in `tsconfig.json`). :-)

sisp:
Do you think it makes sense to add a comment in the code that justifies this "hack" and references #15 for referring to the discussion about the problem?

xaviergonz:
Makes total sense, added, will release this as 0.13.1

sisp:
Thanks a lot for your efforts! You're awesome!

PR #32, opened 2019-09-13
Minor fixes in docs on contexts

sisp:
The node instance `this` was missing in `usernameCtx.setComputed(this, () => this.username)` in the docs section about contexts.

I also fixed a few other minor typos.

### DIFF
--- a/packages/site/src/contexts.mdx
+++ b/packages/site/src/contexts.mdx
@@ -10,7 +10,7 @@ route: /contexts
 Contexts serve as a way to share contextual/environmental data deeply across a tree without having to know the exact structure of such tree.
 Think of it as a dependency injection system.
 
-For example, imagine a state where some children need to know the current username to perform certain operations. We could make the children use `getRoot` to access the username, but that means we need a root for unit testing any of the children, therefore we created a dependency from a child to one of its parents.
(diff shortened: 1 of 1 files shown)


xaviergonz:
Thanks! do you think it solves the issue about unit testing?

sisp:
I think so, or at least it greatly improves testability and also makes model reusability much easier.

That said, I'm still trying to see the equivalent of React's component composition through props for `mobx-keystone`. With contexts, it's a bit like using React *only* with React contexts and without props. But I'm not sure whether an exact equivalent is possible at all (or even makes sense for state management).

xaviergonz:
I couldn't find a way either, but I guess it could be said that the props of a model are the model props injected when you make new Model(...)? 🤷‍♂ 
The good thing about contexts vs onChildAttachedTo is that they are faster, since they go up to fetch the value rather than drill down on all children when any of them gets modified.

sisp:
I'll give it some more thought. My gut feeling is that `mobx-keystone` is similar to a React component class (especially the `state` property corresponds to the model props), but there is no equivalent of the `render` method or any of the lifecycle methods that also depend on the props, too. But React's composition comes from essentially nested `render` calls, so if those don't exist in `mobx-keystone`, maybe there is no real equivalent.

PR #33, opened 2019-09-16
Root ref and other goodies

xaviergonz:
## 0.19.0

- Added `setDefaultComputed` and `getProviderNode` to contexts.
- [BREAKING CHANGE] `getChildrenObjects` will now never report interim data objects (`$`).
- Optimizations to `getChildrenObjects` and `onChildAttachedTo`.
- Added `rootRef`s.

### DIFF
--- a/packages/lib/src/context/context.ts
+++ b/packages/lib/src/context/context.ts
@@ -32,6 +32,13 @@ export interface Context<T> {
   get(node: object): T
 
   /**
+   * Gets node that will provide the context value, or undefined
+   * if it comes from the default.
+   * @param node
+   */
+  getProviderNode(node: object): object | undefined
+
+  /**
    * Sets the context value for a given node, effectively making it a provider.
    * @param node
    * @param value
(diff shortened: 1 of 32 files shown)


packages/lib/src/context/context.ts
```diff
@@ -31,6 +31,13 @@ export interface Context<T> {
    */
   get(node: object): T
 
+  /**
+   * Gets node that will provide the context value, or undefined
```
sisp: undefined -> \`undefined\`

packages/lib/src/ref/Ref.ts
```diff
@@ -0,0 +1,63 @@
...
+export abstract class Ref<T extends object> extends Model({
+  /**
+   * Reference id.
+   */
+  id: tProp(typesString),
+}) {
+  protected abstract resolve(): T | undefined
+
+  /**
+   * The object this reference points to, or undefined if the reference is currently invalid.
```
sisp: undefined -> \`undefined\`

packages/site/src/references.mdx
```diff
@@ -12,9 +12,38 @@ This means that, for example, if we wanted to have a list of todos and a selecte
...
+
+Root references are references that can be resolved as long as both the reference and the referenced object live under the same tree, this is, as long as they share a common root.
+
+They are created like this:
+
+```ts
+const myRef = rootRef<T>("some unique model type id", {
+  getId?(target: unknown): string | undefined {
+    // given an object (which could or could not be of the target type)
+    // what is its id? or undefined if it has no id
```
sisp: undefined -> \`undefined\`

packages/site/src/contexts.mdx
```diff
@@ -62,3 +62,4 @@ The returned context object has the following methods:
 - `set(node, value)` - Sets the (static) value a node will provide for itself and its children. Usually called in `onInit`.
 - `setComputed(node, () => value)` - Sets the (computed) value a node will provide for itself and its children. Usually called in `onInit`.
 - `unset(node)` - Make the node no longer provide a context value.
+- `getProviderNode(node)` - Gets the node that will provide the value, or undefined if it will come from the default value.
```
sisp: undefined -> \`undefined\`

PR #42, opened 2019-09-23
tProp syntax sugar

### LINKED ISSUE #39: Model data syntax
I'm a very new to this project and I love pretty everything, except for one thing: the two syntaxes for model data. I prefer to use the one providing optional type-checking (as it may get useful when things break and I have no clue what's going on), but it's pretty verbose. It's more verbose than with MST and that's IMHO a big drawback.

I guess, plain `done: false` can't work, but couldn't there be just `done: tProp(false)` instead of `done: tProp(types.boolean, false)`?

Am I right, you're using the simpler syntax? I guess, with lots of experience, relying on typescript is pretty safe...

xaviergonz:
Fixes #39 

### Syntax sugar for optional primitives with a default value

You can also do `tProp(defaultValue: string | number | boolean)`, which is equivalent to `tProp(types.string|number|boolean, defaultValue)`.
In other words, if you use `tProp(42)`, then the property will be a number that takes the value `42` when undefined.

### DIFF
--- a/packages/site/src/runtimeTypeChecking.mdx
+++ b/packages/site/src/runtimeTypeChecking.mdx
@@ -266,7 +266,7 @@ const sumModelType = types.refinement(types.model<Sum>(Sum), sum => {
 ### Syntax sugar for optional primitives with a default value
 
 You can also do `tProp(defaultValue: string | number | boolean)`, which is equivalent to `tProp(types.string|number|boolean, defaultValue)`.
-In other words, if you use `tProp(42)`, then the property will be a number that takes the value `42` when undefined.
+In other words, if you use `tProp(42)`, then the property will be a number and take the default value `42` when the value on the snapshot / model creation data is `undefined`.
 
 ### `TypeToData<type>`
 
--- a/packages/site/src/runtimeTypeChecking.mdx
+++ b/packages/site/src/runtimeTypeChecking.mdx
@@ -263,6 +263,11 @@ const sumModelType = types.refinement(types.model<Sum>(Sum), sum => {
 })
 ```
 
+### Syntax sugar for optional primitives with a default value
+
+You can also do `tProp(defaultValue: string | number | boolean)`, which is equivalent to `tProp(types.string|number|boolean, defaultValue)`.
+In other words, if you use `tProp(42)`, then the property will be a number that takes the value `42` when undefined.
+
 ### `TypeToData<type>`
 
 It is also possible to get the type represented by a type via `TypeToData`:
--- a/packages/site/src/runtimeTypeChecking.mdx
+++ b/packages/site/src/runtimeTypeChecking.mdx
@@ -81,19 +81,19 @@ const hiType = types.literal("hi") // the string with value "hi"
 const number5Type = types.literal(5) // the number with value 5
 ```
 
-### `types.undefined`
+### `types.undefined` / `undefined`
 
 A type that represents the value `undefined`.
 
-### `types.null`
+### `types.null` / `null`
 
 A type that represents the value `null`.
 
-### `types.boolean`
+### `types.boolean` / `Boolean`
 
 A type that represents any boolean value.
 
-### `types.number`
+### `types.number` / `Number`
 
 A type that represents any number value.
 
@@ -101,7 +101,7 @@ A type that represents any number value.
 
 A type that represents any integer number value.
 
-### `types.string`
+### `types.string` / `String`
 
 A type that represents any string value.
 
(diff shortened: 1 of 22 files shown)


packages/site/src/runtimeTypeChecking.mdx
```diff
@@ -263,6 +263,11 @@ const sumModelType = types.refinement(types.model<Sum>(Sum), sum => {
 })
 ```
 
+### Syntax sugar for optional primitives with a default value
+
+You can also do `tProp(defaultValue: string | number | boolean)`, which is equivalent to `tProp(types.string|number|boolean, defaultValue)`.
+In other words, if you use `tProp(42)`, then the property will be a number that takes the value `42` when undefined.
```
terrysahaidak: I would use something like `then the property will be a number and use `42` as a default if the value is `undefined`, because it got me confused, I've got it only after reading it for 3rd time or so.

xaviergonz:
I was also thinking about supporting these aliases, but I'm not sure how useful they'd be:

```ts
[A, B, C] => types.or(A, B, C)
Boolean => types.boolean
Number => types.number
String => types.String
null => types.null
undefined => types.undefined

tProp([String, null], null) => tProp(types.or(types.string, types.null), null) => tProp(types.maybeNull(types.string), null)

tProp([String, undefined]) => tProp(types.or(types.string, types.undefined)) => tProp(types.maybe(types.string))

```

would they be useful or just confusing?

packages/site/src/runtimeTypeChecking.mdx
xaviergonz: thanks! clarified

terrysahaidak:
`tProp([String, null], null)` this is really weird looking one :)

For the very first time, I didn't get it, but then realized this is the `types.or`, I would use the word `union` here, for me `[Boolean, String] as types.union(types.boolean, types.string)` just makes more sense.

But yeah, I like it.

Here is a random model from my app converted to this optional syntax:

```ts
class EditablePost extends Model({
  title: tProp(''),
  companyName: '',
  location: tProp([Location, undefined]),
  requirements: tProp(types.array(t.string)),
  startDate: tProp([StringDate, undefined]),
  description: tProp([t.string, undefined]),
  rate: tProp([[t.number, t.string], undefined]),
  hoursPerWeek: tProp([[t.number, t.string], undefined]),
}) {}
```

xaviergonz:
Actually I think it would be more like

```ts
class EditablePost extends Model({
  title: tProp(''),
  companyName: tProp(''),
  location: tProp([t.model<Location>(Location), undefined]),
  requirements: tProp(t.array(String)),
  startDate: tProp([t.model<StringDate>(StringDate), undefined]),
  description: tProp([String, undefined]),
  rate: tProp([[Number, String], undefined]),
  hoursPerWeek: tProp([Number, String], undefined),
}) {}
```

?

terrysahaidak:
Missed the Location, yeah (it was MST model :) ).

But StringDate is a custom type here. This one is more accurate.

```ts
const maybe = [undefined, null];

class EditablePost extends Model({
  title: tProp(''),
  companyName: tProp(''),
  location: tProp([t.model<Location>(Location), undefined]),
  requirements: tProp(t.array(String)),
  startDate: tProp([StringDate, undefined]),
  description: tProp([String, maybe]),
  rate: tProp([[Number, String], maybe]),
  hoursPerWeek: tProp([Number, String], maybe),
}) {}
```

Here is without `tProp`, just for example:

```ts
class EditablePost extends Model({
  title: '',
  companyName: '',
  location: [t.model<Location>(Location), undefined],
  requirements: t.array(String),
  startDate: [StringDate, undefined],
  description: [String, maybe],
  rate: [[Number, String], maybe],
  hoursPerWeek: [[Number, String], maybe],
}) {}
```

xaviergonz:
just wondering, what is `maybe`?

terrysahaidak:
I'm receiving `null` from the server, but my form engine doesn't really play well with `null`, to I might set it as `undefined` at some point. So this is just:

```ts
const maybe = [undefined, null];

// in mst its 
const maybe = types.union(types.undefined, types.null);
```

xaviergonz:
Boolean and friends are done, however they array alias for "or" might require https://github.com/microsoft/TypeScript/pull/33050 (supposedly coming in ts 3.7)

terrysahaidak:
Looking really promising! I think it makes sense to merge it now and add `or` after TS 3.7 reach stable and you will be sure it worth it to set the 3.7 as the required version of TS.

xaviergonz:
@terrysahaidak 
Sounds good :)

PR #43, opened 2019-09-23
add findParentWithPath

xaviergonz:


sisp:
Any particular reason why you named this function `findParentWithPath` and not `findParentPath`? I thought the latter would be more consistent with `getParent`/`getParentPath` (=> `findParent`/`findParentPath`).

xaviergonz:
findParentPath made me think of finding the parent and then its path (like my first comment to your suggestion)

sisp:
I see. For the sake of consistency, would you say it makes sense to rename `getParentPath` to `getParentWithPath` then?

xaviergonz:
Ok, you convinced me, findParentPath it is :)

PR #48, opened 2019-09-26
propTransform as alternative to custom types

xaviergonz:
```ts
/**
 * Prop transform for ISO date strings to Date objects
 * and vice-versa.
 */
export const stringAsDate = propTransform<string | null | undefined, Date | null | undefined>({
  propToData(prop) {
    if (prop == null) return prop
    return new Date(prop)
  },
  dataToProp(date) {
    if (date == null) return date
    return date.toJSON()
  },
})

/**
 * Prop transform for number timestamps to Date objects
 * and vice-versa.
 */
export const timestampAsDate = propTransform<number | null | undefined, Date | null | undefined>({
  propToData(prop) {
    if (prop == null) return prop
    return new Date(prop)
  },
  dataToProp(date) {
    if (date == null) return date
    return +date
  },
})
```

Usage:
```ts
class M extends Model({
  timestamp: prop<number>(),
  dateString: prop<string>(),
}) {
  @timestampAsDate("timestamp")
  date1!: Date

  @stringAsDate("dateString")
  date2!: Date

  @modelAction
  setDate(d: Date) {
    this.date1 = d;
    this.date2 = d;
  }
}

// these are computed props that can be observed, reacted over, etc
m.date1
m.date2
```


xaviergonz:
Will be released in 0.23.0

PR #50, opened 2019-09-27
Models with ids, improved action serialization support

xaviergonz:
This is a big one:

- [BREAKING CHANGE] Added a `$modelId` extra property to models and their snapshots in order to be able to properly validate targets of serialized actions in scenarions with concurrent clients. Also allows the automatic optimization of the serialized version of models in action parameters by substituting them to just their paths + path of ids whenever possible.
- [BREAKING CHANGE] `serializeActionCall`, `serializeActionCallArgument`, `deserializeActionCall` and `deserializeActionCallArgument` now require a first parameter with the root node of the model where actions are going to be performed.
- `ActionContext` now includes `targetPathIds`.
- Added `pathObjects` to `getRootPath`.
- Improved a bit the typing for `fromSnapshot` model methods.
- Improved action argument serialization so it supports dates, maps, sets, arrays of models, etc.

With this change, when synchronizing actions between computers via `serializeActionCall`, `deserializeActionCall` and `applyAction`, the system will automatically detect if an action parameter is a model, and if the model is already on the root store it will send a path reference (validated via ids), rather than the whole object snapshot. If the model is not in the same root store it will just send the snapshot.

Additionally there is better support for serialization of actions that have Maps, Sets, Arrays, etc as arguments.

The new $modelId is basically like a "memory reference" that can be distributed between computers (although it could also have its uses if used as reference ID I guess).


terrysahaidak:
So now each model will have its own unique ID stored in the snapshot and generated in runtime when creating models? Will they be different each time I'm running the app? Will I be able to restore the app from some serialized snapshot (persist)?

xaviergonz:
They will be auto-generated every time you do `new M({....})` (except if you do `new M({[modelId]: "whatever", ....})`, then it will use that one).

They will be persisted in snapshots in `$modelId: "..."`, and restored when using `fromSnapshot`.

If clone is used it will auto-generate new IDs for every model in the tree, to make sure the copy has unique ids.

terrysahaidak:
By the way, nanoid breaks compatibility with React Native cause it uses async version.

xaviergonz:
Hmm the thing about uuid is that it is way slower. Didn't know that nanoid required a different version per environment. I could make it configurable, but I'd like at least to give a good default. Do you have any suggestion faster than uuid that's still unique across machines?

terrysahaidak:
That's actually what I mean: add the ability to configure it per project/env. Nanoid can be a default one, but not sure should it be a dependency or a peer dependency.

Actually don't know any good alternative. Nanoid is good, but it requires crypto module which doesn't exists in RN, you have to use a third-party native module, but each module call going through serializable async bridge to the native side. Maybe it worth to do some benchmarks? According to the author notes, it's only 16% faster: https://www.npmjs.com/package/nanoid#benchmark

Maaartinus:
AFAIK, all you need a single globally unique string (which you can generate e.g., using nanoid) and a counter (also base64-encoded, if you want to keep it very small). This is most probably the fastest method.

An disadvantage is that it leaks information about what IDs were generated in the same run of a client. This probably doesn't matter for most users, so it could be the default.

xaviergonz:
@terrysahaidak In this case it was using the non secure version, which is around 5x faster (maybe that one works in react native out of the box?)

@Maaartinus that sounds feasible, it could use uuid with a number appended at the beginning, the ID won't be pretty but it really should be fast (the ID is only used for reconciliation anyway and serialization anyway and I'll make it configurable just in case anyway)

terrysahaidak:
Seems like UUID (especially the most popular v4 iteration) [uses](https://github.com/kelektiv/node-uuid/blob/master/lib/rng-browser.js) non-secure `Math.random`. The chance of collision exists but I'm not sure it's supercritical, at least in my cases (when optimistically creating some models with temporal id). 

@Maaartinus that would work only if we generate that id each time we initialize our application. Because if it's static, it will break the app if we resolve something from snapshot.

xaviergonz:
I think what he means is basically something like this

```ts
let localID = 0
const baseLocalId = uuid.v4()

function genId() {
  return `${localId++}-${baseLocalId}`
}
```

that way the base id is unique per "application instance", and every application instance generates its own set of unique ids based on that

terrysahaidak:
That might work. Also, if you provide a way to set onw `baseLocalId`, people might even setup secure async nanoid and just set the really secure id before bootstrapping the app.

xaviergonz:
I think I'll just go for something like that by default (using uuidv4 as base) + totally configurable function via "setGlobalConfig" (thought the async base might need to be resolved before calling that, since async id generation is not supported)

terrysahaidak:
If you provide the ability to set that `generateId` function, we can create our own implementation in userland like:

```
let localID = 0
const baseLocalId = await nanoid();

function genId() {
  return `${localId++}-${baseLocalId}`
}

setGlobalConfig({ generateId: genId });
```

xaviergonz:
> If you provide the ability to set that `generateId` function, we can create our own implementation in userland like:
> 
> ```js
> let localID = 0
> const baseLocalId = await nanoid();
> 
> function genId() {
>   return `${localId++}-${baseLocalId}`
> }
> 
> setGlobalConfig({ generateId: genId });
> ```

Exactly :)

xaviergonz:
I think this PR should be ready to go

packages/lib/src/globalConfig/globalConfig.ts
```diff
@@ -1,3 +1,4 @@
+import { v4 as uuidv4 } from "uuid"
```
terrysahaidak: it's better to use 'uuid/v4` or you have a proper setup of code-splitting?

xaviergonz: nice catch, will change it on another pr

PR #72, opened 2019-12-15
Require Typescript 3.7 to support new assertion syntax

sisp:
In 575bac559f3065a6fd86cabf8f2b153649bb1e7e the new assertion syntax introduced in TS 3.7 is used, but `packages/lib/package.json` still requires only [`^3.6.4`](https://github.com/xaviergonz/mobx-keystone/blob/376aa2df28b3272a0652a6ba545e70066f279e67/packages/lib/package.json#L56). The error hasn't surfaced because TS 3.7 is locked in `yarn.lock`.


xaviergonz:
Thanks! Merging.

PR #82, opened 2020-01-01
draft function

xaviergonz:


xaviergonz:
@sisp different (complementary?) to the draft middleware, since it covers a different use case, but I think the name was too good to let it pass. Any suggestions?

sisp:
First of all, this is a great new feature for `mobx-keystone`! The API of `Draft` looks very similar to `mobx-utils`'s [`ViewModel`](https://github.com/mobxjs/mobx-utils#viewmodel) but also works for entire subtrees whereas I believe `ViewModel` can only be used on a flat MobX model.

> [...] (I'm open for new names though)

I think `draft` fits quite well given what it does.

> @sisp different (complementary?) to the draft middleware, since it covers a different use case

Yes, I think the two are complementary. The draft function creates an explicitly managed draft copy of a subtree with methods to manually commit changes to the original subtree or (partially) reset the copy. The draft middleware creates an implicitly managed draft copy of a subtree which is automatically kept in sync with the original subtree and allows to make temporary changes to the draft which must be committed or rejected when the callback returns, so the draft and the original subtree can never diverge. Before nested `withDraft` calls, I think the draft function could have been a replacement for the draft middleware while supporting a broader set of use cases like the preferences form example, which I believe cannot be implemented using the draft middleware because the draft cannot diverge from the original subtree. However, I currently don't see a way to implement nested `withDraft` calls using the draft function because changes to the draft must be associated with the nesting level and accordingly, but the concept of nested drafts does not exist in the draft function.

In the app that I'm building and from which I'm deriving requirements, I need to test "what-if" scenarios which affect the way I visualize UI elements. At this point, I compute some value on the draft after performing some actions but always reject the changes made to the draft in each (nested) `withDraft` call. TBH, I'm still trying to figure out a real world use case for committing changes made in `withDraft`. Could this be the key difference between the draft function and draft middleware, i.e. the draft middleware is only for testing "what-if" scenarios and changes are never committed?

packages/site/src/drafts.mdx
```diff
@@ -0,0 +1,123 @@
...
+  data={preferencesDraft.data}
+  onSave={() => preferencesDraft.commit()}
+  onReset={() => preferencesDraft.reset()}
+  showValidationErrors={preferencesDraft.isDirty}
+  saveDisabled={!preferencesDraft.isDirty || preferencesDraft.data.hasValidationErrors}
+  resetDisabled={!preferencesDraft.isDirty}
+/>
+```
+
+Alternativelly we could pass the draft itself and let the component use the draft properties / methods internally:
```
sisp: Alternativelly -> Alternatively

packages/lib/src/treeUtils/draft.ts
```diff
@@ -0,0 +1,176 @@
...
+import { applyPatches } from "../patch/applyPatches"
+import { applySnapshot } from "../snapshot/applySnapshot"
+import { fromSnapshot } from "../snapshot/fromSnapshot"
+import { getSnapshot } from "../snapshot/getSnapshot"
+import { assertTweakedObject } from "../tweaker/core"
+import { failure } from "../utils"
+import { deepEquals } from "./deepEquals"
+
+/**
+ * A class that with draft related methods.
```
sisp: ~~that~~

packages/lib/src/treeUtils/draft.ts
```diff
@@ -0,0 +1,176 @@
...
+ * Use `draft` to create an instance of this class.
+ *
+ * @typeparam T Data type.
+ */
+export class Draft<T extends object> {
+  /**
+   * Draft data object.
+   */
+  readonly data!: T
```
sisp: Is `!` actually needed? `this.data` is set in the constructor.

xaviergonz: fixed

packages/lib/src/treeUtils/draft.ts
xaviergonz: fixed

PR #107, opened 2020-01-19
maybe fix #96

### LINKED ISSUE #96: Sandbox "tried to invoke action '...' over a readonly node" when original tree contains reaction
I've discovered a problem when the [sandbox](https://mobx-keystone.netlify.com/sandboxes) copy of a node contains a reaction. When the reaction in the original node fires and triggers an action, the sandbox attempts to apply the changes made by the action to the sandbox copy, but it appears the patches are not applied inside `allowWrite(() => { ... }`):

> [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Reaction@15]' MobxKeystoneError: tried to invoke action 'inc' over a readonly node

I believe the problem is caused by this `onPatches` listener:

https://github.com/xaviergonz/mobx-keystone/blob/7ff26354be93b5b5e3796e3a7c1773f3e03c2215/packages/lib/src/treeUtils/sandbox.ts#L79-L81

This is a test case (to be added to [`packages/lib/test/treeUtils/sandbox.test.ts`](https://github.com/xaviergonz/mobx-keystone/blob/master/packages/lib/test/treeUtils/sandbox.test.ts)) that reproduces the problem:

```ts
test("sandbox is patched when original tree with reaction changes", () => {
  @model("R")
  class R extends Model({ a: prop<A>(), n: prop<number>(0) }) {
    onInit() {
      autoDispose(
        reaction(
          () => this.a.b.value,
          () => {
            this.inc()
          }
        )
      )
    }

    @modelAction
    inc() {
      this.n++
    }
  }

  const r = new R({ a: new A({ b: new B({ value: 0 }) }) })
  const manager = sandbox(r)
  autoDispose(() => manager.dispose())

  r.a.b.setValue(10)

  expect(r.a.b.value).toBe(10)
  expect(r.n).toBe(1)

  manager.withSandbox(r, rcopy => {
    expect(rcopy.a.b.value).toBe(10)
    expect(rcopy.n).toBe(1)
    return false
  })
})
```

I suspect that the problem is related to the fact that a reaction is executed after an action has finished and the `onPatches` listener is wrapped as an action. In the code snippet I referenced above, `this.allowWrite` is called inside the `onPatches` listener, so if the reaction is executed after the listener has finished, it means the reaction is executed outside `this.allowWrite`. However, I tried removing

https://github.com/xaviergonz/mobx-keystone/blob/7ff26354be93b5b5e3796e3a7c1773f3e03c2215/packages/lib/src/patch/emitPatch.ts#L59-L61

to prevent the listener from becoming an action and the problem was still present. At this point, I'm not sure what exactly is causing this error.

xaviergonz:
This kind of fixes #96. The problem with this approach though is that this actually allows the sandbox copy and the original copy to be out of sync (as shown in the unit test)

### DIFF
--- a/packages/lib/src/treeUtils/sandbox.ts
+++ b/packages/lib/src/treeUtils/sandbox.ts
@@ -1,4 +1,5 @@
 import { reaction, runInAction } from "mobx"
+import { runAfterCurrentActions } from "../action/runAfterCurrentActions"
 import {
   readonlyMiddleware,
   ReadonlyMiddlewareReturn,
@@ -41,10 +42,10 @@ export class SandboxManager {
   private withSandboxPatchRecorder: PatchRecorder | undefined
 
   /**
-   * Function from `readonlyMiddleware` that will allow actions to be started inside the provided
+   * `readonlyMiddleware` that will allow actions to be started inside the provided
    * code block on a readonly node.
    */
-  private allowWrite: ReadonlyMiddlewareReturn["allowWrite"]
+  private readonlyMw: ReadonlyMiddlewareReturn
 
   /**
    * Creates an instance of `SandboxManager`.
@@ -76,18 +77,22 @@ export class SandboxManager {
       if (this.withSandboxPatchRecorder) {
         throw failure("original subtree must not change while 'withSandbox' executes")
       }
-      this.allowWrite(() => {
-        applyPatches(this.subtreeRootClone, patches)
+
+      this.readonlyMw.writeAllowed = true
+      applyPatches(this.subtreeRootClone, patches)
+
+      // ensure we only stop the write lock until current actions and triggered reactions are finished
+      runAfterCurrentActions(() => {
+        this.readonlyMw.writeAllowed = false
       })
     })
 
-    const { allowWrite, dispose: disposeReadonlyMW } = readonlyMiddleware(this.subtreeRootClone)
-    this.allowWrite = allowWrite
+    this.readonlyMw = readonlyMiddleware(this.subtreeRootClone)
 
     this.disposer = () => {
       disposeReactionRS()
       disposeOnPatches()
-      disposeReadonlyMW()
+      this.readonlyMw.dispose()
       if (isRootStore(this.subtreeRootClone)) {
         unregisterRootStore(this.subtreeRootClone)
       }
@@ -119,7 +124,7 @@ export class SandboxManager {
 
     let commit = false
     try {
-      const returnValue = this.allowWrite(() =>
+      const returnValue = this.readonlyMw.allowWrite(() =>
         fn((isNodesArray ? sandboxNodes : sandboxNodes[0]) as T)
       )
       if (typeof returnValue === "boolean") {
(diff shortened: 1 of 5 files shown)


packages/lib/src/treeUtils/sandbox.ts
```diff
@@ -76,18 +77,22 @@ export class SandboxManager {
...
       }
-      this.allowWrite(() => {
-        applyPatches(this.subtreeRootClone, patches)
+
+      this.readonlyMw.writeAllowed = true
+      applyPatches(this.subtreeRootClone, patches)
+
+      // ensure we only stop the write lock until current actions and triggered reactions are finished
+      runAfterCurrentActions(() => {
+        this.readonlyMw.writeAllowed = false
```
sisp: Would it be better/safer to do the following similar to `allowWrite`?

```ts
const writeAllowedOld = this.readonlyMw.writeAllowed
this.readonlyMw.writeAllowed = true
applyPatches(this.subtreeRootClone, patches)

// ensure we only stop the write lock until current actions and triggered reactions are finished
runAfterCurrentActions(() => {
  this.readonlyMw.writeAllowed = writeAllowedOld
})
```

xaviergonz: Thanks, fixed it in a even more secure manner.

I still feel torn about this though since in this case the exception that readonly middleware throws actually saves you from a reaction that makes the sandbox and the original node get out of sync :-/

sisp: > Thanks, fixed it in a even more secure manner.

Looks even better! :-)

> I still feel torn about this though since in this case the exception that readonly middleware throws actually saves you from a reaction that makes the sandbox and the original node get out of sync :-/

I know what you mean ... In case you didn't see it yet, [@mweststrate doesn't even seem to like the whole sandbox idea](https://github.com/mobxjs/mobx/issues/2259#issuecomment-576199561), but I still believe that it's the only high-performance way to test "what-if" scenarios. I think the costly part is tweaking nodes during `fromSnapshot`, perhaps setting up observers etc., so the more robust `clone` every time a copy is needed, just won't work if this needs to be done at several places for each animation frame.

sisp: @mweststrate's comment about not using reactions to update state

> 1. As documented, updating state from reactions is generally an anti pattern, probably you need computeds, which wouldn't suffer from this idea

got me thinking and perhaps I actually need to change my state design to use a computed property where I'm currently using a reaction.

I'd say, don't merge this PR for now because I share your skepticism about this solution and maybe it's my "un-MobX" way of doing things with a reaction in this case, which shouldn't affect `mobx-keystone`.

sisp:
I think this PR can be closed. I haven't experienced the problem that led to this PR for quite some time and you weren't completely happy with the solution anyway.

xaviergonz:
> I think this PR can be closed. I haven't experienced the problem that led to this PR for quite some time and you weren't completely happy with the solution anyway.

Good to know it doesn't happen anymore :D closing then.

PR #114, opened 2020-01-29
give a way to check if a node is sandboxed

xaviergonz:


packages/lib/test/treeUtils/sandbox.test.ts
```diff
@@ -342,3 +345,103 @@ test("sandbox cannot be changed outside of fn", () => {
...
+      return isSandboxedNode(this)
+    }
+
+    @modelAction
+    shouldBeSandboxed(shouldBeSandboxed: boolean): void {
+      expect(isSandboxedNode(this)).toBe(shouldBeSandboxed)
+    }
+  }
+
+  const a = new A2({ b: new B2({ value: 0 }) })
```
sisp: Should `a` be unregistered as a root store after the test has completed?

```ts
autoDispose(() => {
  if (isRootStore(a)) {
    unregisterRootStore(a)
  }
})
```

sisp:
For some reason, `isSandboxedNode` recognizes `ref` to be a sandboxed node while `next` not to be a sandboxed node:

```ts
test("isSandboxedNode recognizes ref/prev/next to all be sandboxed nodes or not sandboxed nodes", () => {
  const aRef = customRef<A>("aRef", {
    resolve(ref: Ref<A>): A | undefined {
      const maybeR = getParent<R>(ref)
      return maybeR ? maybeR.a : undefined
    },
    onResolvedValueChange(ref, prev, next) {
      if (prev) {
        expect(isSandboxedNode(ref)).toBe(isSandboxedNode(prev))
      }
      if (next) {
        expect(isSandboxedNode(ref)).toBe(isSandboxedNode(next))
      }
    },
  })

  @model("R3")
  class R extends Model({ a: prop<A | undefined>(), aref: prop<Ref<A> | undefined>() }) {}

  const a = new A({ b: new B({ value: 0 }) })
  const r = new R({})
  const manager = sandbox(r)
  autoDispose(() => manager.dispose())

  runUnprotected(() => {
    r.aref = aRef(a)
  })

  runUnprotected(() => {
    r.a = a
  })

  runUnprotected(() => {
    r.a = undefined
  })
})
```

sisp:
And related to the refs problem above, isn't it a problem to create a `computedIdTree` to retrieve a node by its ID when we are dealing with cloned trees with identical IDs?

https://github.com/xaviergonz/mobx-keystone/blob/73ad7dbfe7c952c6ce1f33d539391e4b3e1f6d04/packages/lib/src/ref/rootRef.ts#L75-L76

packages/lib/test/treeUtils/sandbox.test.ts
xaviergonz: done

xaviergonz:
> Edit: Nevermind, it's because a has been detached when onResolvedValueChange is called and, therefore, isSandboxedNode(prev) is false even though it was a sandbox node.

Yep, added the test (with that fixed) anyway :)

xaviergonz:
> Edit: Or maybe not ... In const idMap = computedIdTree!.walk(refRoot), refRoot should be the root node of the cloned tree.

Exactly, that's why they are called "rootRefs" :)

PR #118, opened 2020-02-04
decoratedModel

xaviergonz:
Allows the usage of the lib without decorators

Example

```ts
  const Point = decoratedModel(
    "decoratedModel/Point",
    // although class could be anonymous, by giving it a name 
    // type declarations are shorter
    // and the IDE shows better typings if we do
    class Point extends Model({
      x: prop<number>(),
      y: prop<number>(),
    }) {
      setX(x: number) {
        this.x = x
      }

      setY = (y: number) => {
        this.y = y
      }

      setXY(x: number, y: number) {
        this.setX(x)
        this.setY(y)
      }

      get length() {
        return this.x + this.y
      }

      volatile = "volatile"
    },
    {
      setX: modelAction,
      setY: modelAction,
      setXY: modelAction,
      length: computed,
      // multiple decorators could be applied with xxx: [deco1, deco2...]
    }
  )
  type Point = InstanceType<typeof Point> // needed to be able to do SnapshotInOf<Point>, etc

  const p = new Point<number>({ x: 10, y: 20 })
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 0.39.0
+
+- Added `decoratedModel` so the library can be used without decorators.
+
 ## 0.38.0
 
 - Fixes for `undoMiddleware`, where it wouldn't properly record changes outside a subaction after a subaction is performed.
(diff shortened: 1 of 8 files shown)


xaviergonz:
missing: docs

PR #119, opened 2020-02-04
{WIP] fix a possible typing issude for SnapshotIn when using subclassing

xaviergonz:
not yet sure if needed, needs some investigation


xaviergonz:
Merged as part of decoratedModel, since it was required for it to work ok

PR #196, opened 2020-12-13
Apply and enforce Prettier / linting

sisp:
I've applied Prettier to the entire repo and added Yarn scripts for linting (Prettier for everything and `tsdx lint` for the `lib` package) and fixing Prettier linting errors as well as CI jobs to check for linting errors.

I'm not sure whether you're happy with [Prettier's new default setting for `arrowParens` being `always` since v2.0](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev) because there currently are no parentheses in most cases. If you prefer no parentheses, I can change the config accordingly to override Prettier's default setting.


xaviergonz:
Thanks!

> I'm not sure whether you're happy with Prettier's new default setting for arrowParens being always since v2.0 because there currently are no parentheses in most cases. If you prefer no parentheses, I can change the config accordingly to override Prettier's default setting.

It was like that since that was the default in prettier 1, but both ways are ok for me (either go for the 2.0 default or fix the config to use the old one).

Merged!

sisp:
I'm fine with both, too. Since you've already merged it, I'd say we stick with Prettier's new default setting.

PR #197, opened 2020-12-13
Suppress false positive ESLint warnings

sisp:
I've suppressed false positive ESLint warnings.

### DIFF
--- a/packages/lib/src/typeChecking/enum.ts
+++ b/packages/lib/src/typeChecking/enum.ts
@@ -21,7 +21,12 @@ function enumValues(e: any): (string | number)[] {
  * @ignore
  * Extract enum values out of a enum object.
  */
-export type EnumValues<E> = E extends Record<any, infer V> ? V : never
+export type EnumValues<E> = E extends Record<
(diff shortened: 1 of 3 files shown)


packages/lib/src/typeChecking/enum.ts
```diff
@@ -21,7 +21,7 @@ function enumValues(e: any): (string | number)[] {
  * @ignore
  * Extract enum values out of a enum object.
  */
-export type EnumValues<E> = E extends Record<infer _K, infer V> ? V : never
+export type EnumValues<E> = E extends Record<any, infer V> ? V : never
```
xaviergonz, line 24: I remember I had to actually change any to infer the parameter for enum inference to work correctly, so this warning should be skip instead

sisp: Yes, just realized that, too. I've just pushed another commit that suppresses this warning. It makes the line a bit harder to read. Not sure whether it's worth making ESLint happy at the cost of a little readability loss.

xaviergonz: I think it is OK as long as it doesn't happen often 👍😊

PR #203, opened 2020-12-24
Fix TypeError path of refinement type

sisp:
A refinement type error always had the path `[]` even when the refinement type was used as a child of another type, e.g. `types.object`. I've added a test and fixed the refinement type to use the path passed to the check function instead.


sisp:
By the way, would you like me to add an item in `CHANGELOG.md` too, or do you prefer to manage the changelog yourself?

xaviergonz:
Nice catch! I'll add the entry to the changelog for this one so it is not forgotten, but feel free to add entries for future ones! 👍

PR #210, opened 2021-02-01
Fix grouping sandbox commit patches in a single undo item

sisp:
When committing multiple changes made in a sandbox to the original tree, multiple items in the undo manager's queue were created so that a single call to `undoManager.undo()` did not undo the commit. I think this behavior is not intuitive and unintended.

I've fixed the sandbox to apply the commit patches in a single action, so a single undo item is created for the entire commit.

### DIFF
--- a/packages/lib/test/treeUtils/sandbox.test.ts
+++ b/packages/lib/test/treeUtils/sandbox.test.ts
@@ -16,6 +16,7 @@ import {
   runUnprotected,
   sandbox,
   SandboxManager,
+  undoMiddleware,
   unregisterRootStore,
 } from "../../src"
 import "../commonSetup"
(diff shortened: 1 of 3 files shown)


packages/lib/test/treeUtils/sandbox.test.ts
```diff
@@ -492,3 +493,21 @@ test("isSandboxedNode recognizes ref/prev/next to all be sandboxed nodes or not
...
+  expect(undoManager.redoLevels).toBe(0)
+
+  sandboxManager.withSandbox(b, (node) => {
+    node.setValue(1)
+    node.setValue(2)
+    return { return: undefined, commit: true }
+  })
+
+  expect(undoManager.undoLevels).toBe(1)
+  expect(undoManager.redoLevels).toBe(0)
```
xaviergonz, line 512: could we add here a `expect(undoManager.undoQueue).toMatchInlineSnapshot(...)` just to make sure the recorded patch is ok?

sisp: Good idea, done in 83437a9.

PR #219, opened 2021-02-23
Fix committing non-idempotent sandbox changes to the original subtree

sisp:
I discovered a bug where committing non-idempotent changes in the sandbox to the original subtree resulted in an incorrect sandbox state. The problem was that applying the sandbox patches to the original subtree triggered the `onPatches` listener that applies changes in the original subtree to the sandbox to keep them in sync. When committing changes in the sandbox to the original subtree, the patches emitted in the original subtree should be ignored by the sandbox.

This PR fixes the bug.

### DIFF
--- a/packages/lib/src/treeUtils/sandbox.ts
+++ b/packages/lib/src/treeUtils/sandbox.ts
@@ -74,6 +74,11 @@ export class SandboxManager {
   private allowWrite: ReadonlyMiddlewareReturn["allowWrite"]
 
   /**
+   * Whether changes made in the sandbox are currently being committed to the original subtree.
+   */
+  private isComitting: boolean = false
+
+  /**
    * Creates an instance of `SandboxManager`.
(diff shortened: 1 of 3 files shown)


packages/lib/src/treeUtils/sandbox.ts
```diff
@@ -222,7 +229,9 @@ export class SandboxManager {
           for (let i = 0; i < len; i++) {
             patches.push(...recorder.events[i].patches)
           }
+          this.isComitting = true
```
xaviergonz, line 232: Save the old value then restore the value in a finally block to ensure a failing apply won't break it

xaviergonz: The saving is to make sure if for some reason two calls of this get into the stack it will be restored properly

sisp: Good idea, done in 31869e5.

Issue #233, opened 2021-04-18
Computed backrefs without reaction


sisp:
[Backrefs are currently updated using a `reaction`](https://github.com/xaviergonz/mobx-keystone/blob/0444b0f95bb316d0e3446e2b03d41d6f226ae5e4/packages/lib/src/ref/core.ts#L66-L82) which means a new backref is only available by `getRefsResolvingTo` after the outermost (trans)action has completed. I think it would be better if backrefs did not rely on a `reaction` but instead were computed values.

I think computing backrefs without a `reaction` can be achieved by using an approach similar to root refs based on [`computedWalkTreeAggregate`](https://github.com/xaviergonz/mobx-keystone/blob/0444b0f95bb316d0e3446e2b03d41d6f226ae5e4/packages/lib/src/parent/walkTree.ts#L138-L161). For root refs,  `computedWalkTreeAggregate` walks the tree and creates a reactive computed map of IDs and the corresponding nodes by, so that a root ref can access this map to resolve a reference as long as the node and the ref are part of the same tree. Backrefs could be computed in a similar way, i.e. a map of IDs and and refs that resolve the corresponding nodes could be created. While the map created for root refs contains 1:1 mappings of IDs and nodes, the map needed for backrefs would contain 1:n mappings of IDs and nodes because multiple refs can resolve the same node. Perhaps `computedWalkTreeAggregate` could be generalized so that the map values can be arrays, essentially implementing a "group-by" operation when the predicate returns the same value for multiple nodes.

xaviergonz:
I assume you want to do this to be able to access updated back references from the middle of an action. I'm not sure if cached computeds return a stale value when invoked from actions or not. Btw, would it be OK for your use case to trigger onResolvedValueChanged inside a reaction?

sisp:
Correct. Regarding `onResolvedValueChanged`, ideally no reaction would be used here either, but this is trickier I think (https://github.com/xaviergonz/mobx-keystone/issues/214#issuecomment-792839090, https://github.com/mobxjs/mobx/issues/2795, https://github.com/mobxjs/mobx/pull/2802), but perhaps you have an idea.

sisp:
> I'm not sure if cached computeds return a stale value when invoked from actions or not (would need to test).

This test passes:

```ts
test("cached computed returns an up-to-date value when invoked from an action", () => {
  const a = observable.box(1)
  const b = computed(() => a.get() * 2)

  expect(b.get()).toBe(2)
  
[...]

PR #234, opened 2021-04-22
Force update back refs

xaviergonz:


packages/lib/src/ref/core.ts
```diff
@@ -154,35 +182,61 @@ function getBackRefs(target: object, refType?: RefConstructor<object>): Observab
...
+  }
 ): ObservableSet<Ref<T>> {
   assertTweakedObject(target, "target")
 
   const refTypeObject = refType as RefConstructor<object> | undefined
+
+  if (options?.updateAllReferencesIfNeeded && isReactionDelayed()) {
+    // in this case the reference update might have been delayed
+    // so we will make a best effort to update them
+    allRefs.forEach((ref) => ref.forceUpdateBackRefs())
```
sisp, line 203: If `refType` is defined, do you think it's worth calling `ref.forceUpdateBackRefs()` only if `ref instanceof refType`, i.e. is the extra cost of the check worth saving some unnecessary forced updates?

packages/lib/src/ref/core.ts
```diff
@@ -154,35 +182,61 @@ function getBackRefs(target: object, refType?: RefConstructor<object>): Observab
...
+      getBackRefs(oldTarget, refClass as RefConstructor<any>).delete(ref)
+    }
+    if (newTarget) {
+      getBackRefs(newTarget).add(ref)
+      getBackRefs(newTarget, refClass as RefConstructor<any>).add(ref)
+    }
   }
+)
+
+function isReactionDelayed() {
```
sisp, line 232: Really clever! 😍

packages/lib/src/ref/core.ts
sisp, line 203: > I try to avoid it because marking a node as a root store is (in theory) optional, but very practical to be able to get that hook working. [...]
>
> Source: https://github.com/xaviergonz/mobx-keystone/issues/233#issuecomment-825216171

Or instead of maintaining the `allRefs` set via the `onAttachedToRootStore` hook, what about [your other suggestion using `walkTree`](https://github.com/xaviergonz/mobx-keystone/issues/233#issuecomment-823530686)?

```ts
walkTree(fastGetRoot(target), n => {
  if (n instanceof Ref) {
    n.forceUpdateBackRefs()
  }
)
```

It probably is more expensive to walk the tree instead of iterating over just ref nodes, but it would work for trees that are not root stores. Backrefs across trees would still not work though (not my use case at the moment, just saying). Could the performance be improved by using [`computedWalkTreeAggregate`](https://github.com/xaviergonz/mobx-keystone/blob/0444b0f95bb316d0e3446e2b03d41d6f226ae5e4/packages/lib/src/parent/walkTree.ts#L138-L161) due to caching?

packages/lib/src/ref/core.ts
```diff
@@ -154,35 +182,61 @@ function getBackRefs(target: object, refType?: RefConstructor<object>): Observab
...
+ * @param refType Pass it to filter by only references of a given type, or do not to get references of any type.
+ * @param options Options.
  * @returns An observable set with all reference objects that point to the given object.
  */
 export function getRefsResolvingTo<T extends object>(
   target: T,
-  refType?: RefConstructor<T>
+  refType?: RefConstructor<T>,
+  options?: {
+    updateAllReferencesIfNeeded?: boolean
```
sisp, line 193: Perhaps `updateAllRefsIfNeeded` to be more consistent with calling references "refs" everywhere else?

packages/lib/src/ref/core.ts
```diff
@@ -56,6 +62,28 @@ export function internalCustomRef<T extends object>(
...
 
+    onAttachedToRootStore() {
+      allRefs.add(this)
+
+      return () => {
+        allRefs.delete(this)
+      }
+    }
+
+    #savedOldTarget: T | undefined
```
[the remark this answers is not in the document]
xaviergonz: Just wanted to start using it, since they are transpiled to something really private instead of only private if you are using typescript. Do you see any cons?

packages/lib/src/ref/core.ts
```diff
@@ -56,6 +62,28 @@ export function internalCustomRef<T extends object>(
       return this.resolver(this)
     }
 
+    onAttachedToRootStore() {
```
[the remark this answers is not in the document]
xaviergonz: Good catch!

packages/lib/src/ref/core.ts
xaviergonz: Good point!

packages/lib/src/ref/core.ts
xaviergonz: 😊

packages/lib/src/ref/core.ts
xaviergonz: Walk tree would be slower. The computed one would be not as slow, but still slower too.

packages/lib/src/ref/core.ts
xaviergonz: Good point

packages/lib/src/ref/core.ts
sisp: I'd say it's your call. I always register a root store, but the (computed) tree walk would be more consistent with the rest of the library regarding keeping root stores optional.

packages/lib/src/ref/core.ts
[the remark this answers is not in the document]
sisp: No, just wondering. Is there any performance overhead?

packages/lib/src/ref/core.ts
xaviergonz: Even if you call a tree walk, where would you start the walk from?
- From the root of the target model? (the ref might be in another tree)
- From the root of all root stores? (same result as if it used onAttachedToRootStore but slower)

Maybe if the target has a root (but is not part of a root store) use walk tree of the target tree + all root stores to update?
Dunno, I feel it is over complicating for a rare use case of an already rare use case :)

packages/lib/src/ref/core.ts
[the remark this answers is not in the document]
xaviergonz: Very minimal I think (when transpiled)

packages/lib/src/ref/core.ts
sisp: Ah, I forgot that the solution that uses `onAttachedToRootStore` is able to resolve references even across multiple trees.

Looks good then, let's keep it as it is now. :slightly_smiling_face:

sisp:
I think this PR looks great! Thank you very much for implementing this feature! :slightly_smiling_face:

xaviergonz:
@sisp actually I might need to think about this PR a bit more since it introduces a potential mem leak

Say you create a root store, with a reference to some node N
This adds the reference N to the Map, thus N and whatever it points to cannot be garbage collected (alongside anything the model being pointed to points to and so on...)
If we unregister the root store it will be garbage collected, but if we don't (e.g., the root store pointer is "lost") then it won't.

xaviergonz:
I updated it with an strategy based on premise that the target and the back-reference being part of the same root store, so it should work for root references at least without leaking.

sisp:
You're right about the potential memoy leak. But now a tree walk would work, too, wouldn't it? So a root store would not be required at the cost of a more expensive (computed) walk. Just throwing in the tree walk again because you have been avoiding the `onAttachedToRootStore` hook internally.

xaviergonz:
True, I changed the approach so it uses "deepChildren" instead. That way it does not require a root store (though refs must be under the same root as the target when checked inside an action) and also it does not need to walk the whole tree.

sisp:
It looks absolutely fantastic! Thank you so much for your the efforts on optimizing the implementation of this feature!

sisp:
Just one last thing: The docs need a small update about this new behavior and the `updateAllRefsIfNeeded` option.

xaviergonz:
Docs updated

xaviergonz:
out in 0.57.0

sisp:
Thank you very much!

PR #242, opened 2021-05-03
improved generics

xaviergonz:


xaviergonz:
@sisp could you please review it?

sisp:
Looks great in my opinion!

The only thing I noticed is that the generic types must be provided explicitly when instantiating a model, they aren't inferred.

```ts
@model("GenericModel")
class GenericModel<T1, T2> extends Model(<U1, U2>() => ({
  v1: prop<U1>(),
  v2: prop<U2>(),
  v3: prop<number>(0),
}))<T1, T2> {}

const s = new GenericModel({ v1: "1", v2: 2, v3: 3 }) // GenericModel<unknown, unknown>
```

I'm not sure if it's possible to get those generics inferred though.

xaviergonz:
> I'm not sure if it's possible to get those generics inferred though.

No idea, if you find a way let me know. It seems you've really looked into it :)

sisp:
Thank you very much for your efforts on improving generic models! 🙏  This is a great improvement in my opinion. If I find a way to have the generics inferred, I'll let you know.

xaviergonz:
> If I find a way to have the generics inferred, I'll let you know.

Please do, I'm really curious :)

xaviergonz:
Thanks for the pointers. Generics should now be inferable in v0.58.2 :D

sisp:
Incredible, I don't know how long it would have taken me to find that solution! Thank you very much (again), you're a genius! 😃

xaviergonz:
My only skill is to be able to throw random attempts to fix it at the wall for long enough until it works 😉

sisp:
random -> educated guessing

I'd say 😉.

Issue #247, opened 2021-05-10
Get filtered set of nodes in a subtree


sisp:
I'd like to be able to get a _computed_ set of all nodes in a subtree optionally filtered by some predicate. It seems [`computedWalkTreeAggregate`](https://github.com/xaviergonz/mobx-keystone/blob/9be6e0f386547c7b5a58eedd3566bbfa6a63fece/packages/lib/src/parent/walkTree.ts#L138-L161) is already doing the heavy lifting. Would you be open to adding such a function? E.g.:

```ts
function getNodes(subtreeRoot: object, predicate?: (node: object) => boolean): Set<object> {
  // ...
}
```

(I'm open to finding a better name :smile:.)

sisp:
Regarding the name, maybe `findNodes` like [`findParent`](https://github.com/xaviergonz/mobx-keystone/blob/9be6e0f386547c7b5a58eedd3566bbfa6a63fece/packages/lib/src/parent/findParent.ts#L17)?

sisp:
It seems that [`getChildrenObjects`](https://github.com/xaviergonz/mobx-keystone/blob/9be6e0f386547c7b5a58eedd3566bbfa6a63fece/packages/lib/src/parent/getChildrenObjects.ts#L12-L25) with `{ deep: true }` already returns a computed set of non-primitive nodes. Would it be better to use

1. `getChildrenObjects` and filter the returned set using the predicate, or to use
2. `computedWalkTreeAggregate` and convert the returned map to a set?

I _think_ `computedWalkTreeAggregate` might be more efficient because the predicate is applied in each nested computed value, which caches the result, whereas with `getChildrenObjects` any change in the subtree requires filtering the entire set of nodes returned by `getChildrenObjects`. What do you think?

sisp:
I've submitted a PR with an attempt at closing this issue. I went for using `getChildrenObjects` because
1. it's simpler,
1. I think filtering a set should be quite cheap, and
1. `computedWalkTreeAggregate` requires a key for each entry in the returned map which is a model ID for root refs and `resolveId` but the key for e.g. an array node is not so obvious and I think `findChildren` (as a complement to `findParent`) should not be restricted to finding only model nodes.

Let me know what you think :slightly_smiling_face:

xaviergonz:
Thanks for making the feature!

sisp:
You're welcome, thanks for accepting it! 🙂

PR #252, opened 2021-05-12
Add findChildren

sisp:
An attempt at closing #247.

### DIFF
--- a/packages/lib/src/parent/findChildren.ts
+++ b/packages/lib/src/parent/findChildren.ts
@@ -1,0 +1,29 @@
+import { getChildrenObjects } from "./getChildrenObjects"
+
+/**
+ * Iterates through all children and collects them in a set if the
+ * given predicate matches.
+ *
+ * @param root Root object to get the matching children from.
+ * @param predicate Function that will be run for every child of the root object.
+ * @returns A readonly observable set with the matching children.
+ */
+export function findChildren<T extends object = any>(
+  root: object,
+  predicate: (node: object) => boolean
+): ReadonlySet<T> {
+  const children = getChildrenObjects(root, { deep: true })
+
+  const set = new Set<any>()
+
+  const iter = children.values()
+  let cur = iter.next()
+  while (!cur.done) {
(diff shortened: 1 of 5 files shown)


packages/lib/src/parent/findChildren.ts
```diff
@@ -0,0 +1,29 @@
...
+ * given predicate matches.
+ *
+ * @param root Root object to get the matching children from.
+ * @param predicate Function that will be run for every child of the root object.
+ * @returns A readonly observable set with the matching children.
+ */
+export function findChildren<T extends object = any>(
+  root: object,
+  predicate: (node: object) => boolean
+): ReadonlySet<T> {
```
xaviergonz, line 14: maybe add
```
,
  options?: {
    deep?: boolean
  }
```
?

xaviergonz (commented):
hey, thanks for the PR! just left a comment there

packages/lib/src/parent/findChildren.ts
```diff
@@ -0,0 +1,29 @@
...
+  const children = getChildrenObjects(root, { deep: true })
+
+  const set = new Set<any>()
```
xaviergonz, line 17: I think it can be Set<object>

sisp: Then I get:
```
Type 'Set<object>' is not assignable to type 'ReadonlySet<T>'.
  Types of property 'forEach' are incompatible.
    Type '(callbackfn: (value: object, value2: object, set: Set<object>) => void, thisArg?: any) => void' is not assignable to type '(callbackfn: (value: T, value2: T, set: ReadonlySet<T>) => void, thisArg?: any) => void'.
      Types of parameters 'callbackfn' and 'callbackfn' are incompatible.
        Types of parameters 'value' and 'value' are incompatible.
          Type 'object' is not assignable to type 'T'.
            'object' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'object'.ts(2322)
```

xaviergonz:
Thanks!

PR #259, opened 2021-06-03
Implement arrayActions.swap function (as suggestion to close #258)

### LINKED ISSUE #258: A way to extend arrayActions / objectActions ?
Hey @xaviergonz, first: thank you for providing Mobx Keystone, I'm using it in my app and liking it much better than Mobx State Tree! 

Currently I'm trying to figure out a possibility to extend generic array / object actions without tying it up to a particular model instance. 

In concrete:  `arraySwap` function is currently non-generic and requires a "host" model: 
```typescript
@modelAction arraySwap(array: any[], indexA: number, indexB: number) {
    if (indexA < 0 || indexB < 0 || indexA > array.length - 1 || indexB > array.length - 1) {
      return
    }
    var tmp1 = array[indexA]
    var tmp2 = array[indexB]
    array[indexA] = null
    array[indexB] = null
    array[indexA] = tmp2
    array[indexB] = tmp1
    return array
  }
```
When implemented with `arrayActions` primitives, it blows up my React.js app. 

```typescript
class ArrayUtils {
  static arraySwap(array: any[], indexA: number, indexB: number) {
    if (indexA < 0 || indexB < 0 || indexA > array.length - 1 || indexB > array.length - 1) {
      return
    }
    var tmp1 = array[indexA]
    var tmp2 = array[indexB]
    arrayActions.set(array, indexA, null)
    arrayActions.set(array, indexB, null)
    arrayActions.set(array, indexA, tmp2)
    arrayActions.set(array, indexA, tmp1)
    return array
  }
}
```

I see that `arrayActions` is using `import { fnModel } from "./fnModel"`, and `fnModel` is not exposed for end users. 

Are there any other options to implement further generic Array/Object actions, without using a mounted model instance? 

As a short-term fix I could contribute a `swap` function on the `ArrayActions` module, yet this feels rather inflexible.

mindreframer:
Problem: 
- currently it’s not possible to swap elements in an array without relying on a “host” mobx keystone model
- there are currently not user-exposed ways to extend arrayActions

Solution / Workaround: 
- implement a (non-standard) array function for now directly on arrayActions module
- + tests

### DIFF
--- a/packages/lib/src/standardActions/arrayActions.ts
+++ b/packages/lib/src/standardActions/arrayActions.ts
@@ -17,7 +17,17 @@ const _arrayActions = fnModel<unknown[]>("mobx-keystone/arrayActions").actions({
   setLength(length: number): void {
     this.length = length
   },
-
+  swap(indexA, indexB: number): void {
+    if (indexA < 0 || indexB < 0 || indexA > this.length - 1 || indexB > this.length - 1) {
+      return
+    }
+    var tmp1 = this[indexA]
+    var tmp2 = this[indexB]
+    set(this, indexA, null)
+    set(this, indexB, null)
+    set(this, indexA, tmp2)
+    set(this, indexB, tmp1)
+  },
   concat(...items: ConcatArray<any>[]): any[] {
     return this.concat(...items)
   },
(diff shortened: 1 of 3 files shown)


mindreframer:
As continuation to https://github.com/xaviergonz/mobx-keystone/issues/258. Hopefully there is a less invasive way to solve it.

mindreframer:
I would suggest closing this PR in favour of https://github.com/xaviergonz/mobx-keystone/pull/262

PR #269, opened 2021-06-21
Deprecate prop modifier variant withSetter("assign")

sisp:
Clarifies usage of `prop().withSetter("assign")` in the docs. Resolves #268.


xaviergonz:
the assign mode is legacy and deprecated, that's why it is not documented

sisp:
Ah, I didn't know. Should we update the docs stating it's legacy or just leave things as they are?

xaviergonz:
I guess the former so people don't get confused 😅

sisp:
I've updated the PR to deprecate `withSetter("assign")` using the JSDoc tag `@deprecated` and documented the deprecation in the docs. What do you think?

xaviergonz:
Thanks!

PR #318, opened 2021-09-12
make modelId no longer a requirement

xaviergonz:


xaviergonz:
@sisp  @justin-hackin @amorsillo

I know this isn't a full solution to the metadata dilema ($modelType+$modelId), but do you feel this breaking change would help you somehow?

justin-hackin:
@xaviergonz thanks a lot for considering my feedback. This will remote a lot of cruft from my snapshots so I'm happy to see it.  Even without this, I was satisfied with my metadata solution of just storing the context data in a registry based on path from root node in order to preserve metadata passed in defaultFn initialization.  Now I'm wondering, will this change the behavior of applySnapshot? Namely, if a model in a nested tree doesn't have a $modelId, will it still be detached and a new node re-attached?

xaviergonz:
> will this change the behavior of applySnapshot? Namely, if a model in a nested tree doesn't have a $modelId, will it still be detached and a new node re-attached

right now that only happens if the model ids are different. if they are the same the same node gets reused.

after this PR the same keeps happening for models with ids, but for models without ids it will treat the as if they had the same ID (try to reuse the node)

justin-hackin:
> after this PR the same keeps happening for models with ids, but for models without ids it will treat the as if they had the same ID (try to reuse the node)

That's what I suspected, even better!

sisp:
@xaviergonz Looks good to me. Just one question: Am I getting it right that a model without an explicitly set `idProp` property cannot be used with root references? Custom references would still work as long as the referenced model can be resolved somehow, I guess.

xaviergonz:
@sisp both things are absolutely right

justin-hackin:
Eagerly awaiting this vast improvement, I had to do some wacky stuff with recursively deleting $modelId and merging with current snapshot before applying snapshot in order to prevent computed properties created in onAttachedToRootStore from being blown away. I tried to check out the branch in order to test with npm linking but I failed to install dependencies with yarn `YN0066: │ typescript@patch:typescript@npm%3A4.4.3#builtin<compat/typescript>::version=4.4.3&hash=a45b0e: Cannot apply hunk #2 (set enableInlineHunks for details)`

xaviergonz:
Since it has been a few weeks and nobody opposed to it it is now out in 0.62.0

PR #326, opened 2021-09-30
add snapshot processors for class models

xaviergonz:
- [BREAKING CHANGE] The model override `fromSnapshot` has been removed in favor of `fromSnapshotProcessor` / `toSnapshotProcessor` for `Model` / `ExtendedModel`. These allow you to set up pre/post snapshot processors for models.

## Snapshot pre/post-processors: `fromSnapshotProcessor` / `toSnapshotProcessor`

`fromSnapshotProcessor` might be used to transform an input snapshot into the model expected input snapshot.
This is useful for example for versioning.
Note that the `$modelType` property will always be there and will remain unchanged no matter the transformation.

```ts
@model("name")
class Name extends Model(
  {
    // in version 2 we split fullName into firstName and lastName
    _version: prop(2),
    firstName: prop<string>(),
    lastName: prop<string>(),
  },
  {
    fromSnapshotProcessor(
      sn: { _version: 2; firstName: string; lastName: string } | { _version: 1; fullName: string }
    ) {
      if (sn._version === 2) {
        return sn
      }

      const [firstName, lastName] = sn.fullName.split(" ")

      return {
        _version: 2,
        firstName,
        lastName,
      }
    },
  }
) {
  // ...
}
```

`toSnapshotProcessor` is the opposite and might be used to transform the model expected output snapshot into another kind of object snapshot.
Note that the `$modelType` property will always be there and will remain unchanged no matter the transformation.

```ts
@model("name")
class Name extends Model(
  {
    firstName: prop<string>(),
    lastName: prop<string>(),
  },
  {
    toSnapshotProcessor(sn, modelInstance) {
      // we want to also keep fullName for backwards compatibility
      return {
        ...sn,
        fullName: `${sn.firstName} ${sn.lastName}`,
      }
    },
  }
) {
  // ...
}
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,11 @@
     "titleBar.activeBackground": "#68217a",
     "titleBar.inactiveBackground": "#68217a99",
     "titleBar.activeForeground": "#e7e7e7",
-    "titleBar.inactiveForeground": "#e7e7e799"
+    "titleBar.inactiveForeground": "#e7e7e799",
+    "editorGroup.border": "#8a2ca2",
+    "panel.border": "#8a2ca2",
+    "sash.hoverBorder": "#8a2ca2",
+    "sideBar.border": "#8a2ca2"
   },
   "search.exclude": {
     "**/.yarn": true,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.63.0
 
-- [BREAKING CHANGE] The model override `fromSnapshot` has been removed in favor of `fromSnapshotProcessor` / `toSnapshotProcessor` for `Model` / `ExtendedModel`. These allow you to set up pre/post snapshot processors for models.
+- [BREAKING CHANGE] The model override `fromSnapshot` has been removed in favor of `fromSnapshotProcessor` / `toSnapshotProcessor` options of `Model` / `ExtendedModel`. These allow you to set up pre/post snapshot processors for models.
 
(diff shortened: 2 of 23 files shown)


sisp:
Thanks for implementing this feature and also the version example, which I find very helpful!

A comment and a question:

1. I find calling `fromSnapshotProcessor`/`toSnapshotProcessor` a bit hard to read, especially with multiple calls. How about a builder pattern style API?
   ```ts
   class M extends Model(…)
     .fromSnapshotProcessor(…)
     .fromSnapshotProcessor(…) {
     // …
   }
   ```
1. Just to be sure I understand the feature correctly: Both `fromSnapshotProcessor` and `toSnapshotProcessor` can be called on the same model, right? And multiple calls of the same processor behave like chaining the processors?

xaviergonz:
> ```ts
> class M extends Model(…)
>   .fromSnapshotProcessor(…)
>   .fromSnapshotProcessor(…) {
>   // …
> }
> ```

That was my first attempt, but adding builders to model breaks "the new way to make generics" that you devised :(

In the end I think I'm gonna move those functions to the model options, which should be more readable
```ts
class M extends Model(..., {
  fromSnapshotProcessor(...) {},
  toSnapshotProcessor(...) {}
}) {}
```

The only downside is that it doesn't make them directly composable anymore (they are with the original version), but then again you can just call multiple functions inside the fromSnapshotProcessor if needed

sisp:
Ah, I see. Yes, that would be more readable in my opinion. 🙂

PR #347, opened 2021-11-13
migrate docs from Docz to Docusaurus

### LINKED ISSUE #341: Migrate docs from Docz to Docusaurus
The docs are currently based on [Docz](https://www.docz.site/) which is lightweight, quite flexible, and based on Gatsby :+1:. However, I'm thinking it might be nice to switch to [Docusaurus](https://docusaurus.io/) for a couple of reasons:

- Docusaurus sites are really beautiful by default! :heart_eyes:
- Easy theming, e.g. color adjustments according to the project's brand
- Light/Dark mode out of the box
- Full-text [search](https://docusaurus.io/docs/search) (via a search service or also with a local index)
- Site navigation sidebar available out of the box which helps with quick navigation to sections
- Built-in support for [explicit heading IDs](https://docusaurus.io/docs/markdown-features/headings#explicit-ids), e.g. for linking to a section on a page like [https://mobx-keystone.js.org/snapshots#onsnapshot](https://mobx-keystone.js.org/snapshots#onsnapshott-extends-objectobj-t----t-listener-sn-snapshotoutoft-prevsn-snapshotoutoft--void---void) where the auto-generated ID is ugly
- MDX (like Docz)
- Widely-used, e.g. [Create React App](https://create-react-app.dev/), [MobX](https://mobx.js.org/), [mobx-state-tree](https://mobx-state-tree.js.org/intro/welcome), [Immer](https://immerjs.github.io/immer/), ...

What do you think?

sisp:
This PR migrates the documentation from Docz to Docusaurus.

A couple of comments:
- I've added a favicon.
- The examples don't require copies of the source code files with a file suffix `.txt` anymore for showing the code in the documentation.
- I don't use a `<Playground />`-like component in the examples anymore because I think it doesn't add any value to show the source code of only the component call.
- I use [Bootstrap Icons](https://icons.getbootstrap.com/) as an icon library. It's a large collection of beautiful icons, has a permissive license, and the package just contains SVGs that are loaded via [`svgr`](https://react-svgr.com/), so there's no need for any special React icon (wrapper) library, e.g. [`react-fontawesome`](https://github.com/FortAwesome/react-fontawesome) or [`@mui/icons-material`](https://mui.com/components/material-icons/).
- In `patches.mdx`, `snapshots.mdx`, and `treeLikeStructure.mdx`, I've replaced headings with (sometimes) long function signatures by just function names and block code right below containing the function signatures because the headings were difficult to read in the on-page sidebar on the right. What do you think about this change?
- I've kept the URL paths identical to the Docz-based documentation. However, I think it would be a bit nicer to use lower case characters and dashes to separate words. This would break existing links to the docs though unless we find a way to create redirects from the current to the new URLs.
- I've not added a search box yet and would postpone it to a follow-up PR. Would you prefer using the Algolia search service or a local one? For local search, I think https://github.com/cmfcmf/docusaurus-search-local looks like the best option.

Let me know what you think. :slightly_smiling_face:

Closes #341.


xaviergonz:
that looks awesome! 👍I'll take a deeper look as soon as my laptop is repaired 😅

xaviergonz:
Just had the chance to look at it. it is awesome!! ❤️ 

> I've kept the URL paths identical to the Docz-based documentation. However, I think it would be a bit nicer to use lower case characters and dashes to separate words. This would break existing links to the docs though unless we find a way to create redirects from the current to the new URLs.

Feel free to change them

> I've not added a search box yet and would postpone it to a follow-up PR. Would you prefer using the Algolia search service or a local one? For local search, I think https://github.com/cmfcmf/docusaurus-search-local looks like the best option.

I prefer local, but if it is too annoying to setup the algolia would be ok too.

xaviergonz:
By the way, I'm gonna merge it already, but feel free to open another PR to add search / fix urls!

sisp:
Awesome! :slightly_smiling_face: Thanks also for fixing the Netlify config, I had forgotten about it.

sisp:
By the way, what happened to the preview deployments of the docs on Netlify?

dodas:
Thanks for this, it's beautiful!
And looking forward to the full text search!

PR #350, opened 2021-11-19
update CI badge

sisp:
This PR updates the CI badge from Circle CI to GitHub Actions.


sisp:
I just noticed this badge is not flat. :thinking: I'm not sure whether there's a way to set the style for GitHub's own Actions badge. We could use shields.io but might not get the cool GitHub logo in the badge. What do you think?

sisp:
Already figured it out. shields.io supports named logos. I've added a commit for using shields.io instead of GitHub's own Actions badge.

PR #352, opened 2021-11-25
add search to docs

sisp:
This PR adds search to the docs using [`@easyops-cn/docusaurus-search-local`](https://github.com/easyops-cn/docusaurus-search-local).

I looked at all local search solutions for Docusaurus and found this one to be the best (IMO) considering functionality, design, code quality, and maintenance.

- [`docusaurus-plugin-lunr`](https://github.com/daldridge/docusaurus-plugin-lunr) - Appears to be umaintained.
- [`docusaurus-lunr-search`](https://github.com/lelouch77/docusaurus-lunr-search) - [No support for dark mode](https://github.com/lelouch77/docusaurus-lunr-search/issues/55) yet, written in JavaScript (not TypeScript), [requires manual swizzling](https://github.com/lelouch77/docusaurus-lunr-search#how-to-use-), not actually tested. But generally looks not too bad. Might be an alternative.
- [`@cmfcmf/docusaurus-search-local`](https://github.com/cmfcmf/docusaurus-search-local) - Needs some style adjustments, a bit difficult to style (mainly via Algolia CSS variables which are a bit odd because of `-rgb` and `-alpha` variables, so reusing `--ifm-` color varialbes doesn't work), result highlighting sometimes breaks text rendering and cannot be disabled, code quality could be better (IMO), [hiding result highlighting requires page reload or navigating to another page](https://github.com/cmfcmf/docusaurus-search-local/issues/42).
- [`docusaurus-theme-search-typesense`](https://github.com/typesense/docusaurus-theme-search-typesense) - Requires a Typesense server.

On the other hand, [`@easyops-cn/docusaurus-search-local`](https://github.com/easyops-cn/docusaurus-search-local) is easy to set up, requires minimal style customization, supports light and dark mode out of the box, code quality looks quite good (IMO), looks sufficiently maintained (although there's not a ton of activity), has a search page. I probably would have preferred a modal/overlay like Algolia's instead of a popover. Just a minor issue: For very small screens, the popover leaves the viewport at the left of the screen, but that's probably a rare case and could be fixed if needed.

A final note: In order to try the search feature, you need to _build_ the docs and serve the built files, i.e. run:
```bash
yarn site:build
yarn site:serve
```

Search doesn't work with the development server (`yarn site:start`).

Alternatively, we could use Algolia, of course. What do you think?

Follow-up of #347.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,8 +6,8 @@ dist/
 /packages/lib/LICENSE
 /packages/lib/CHANGELOG.md
 
-/packages/site/.docz/
-/packages/site/public/api/
+/packages/site/.docusaurus/
+/packages/site/build/
 
 .yarn/*
 !.yarn/patches
--- a/package.json
+++ b/package.json
@@ -28,5 +28,9 @@
     "react-scripts": "^4.0.3",
     "typescript": "^4.5.2"
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
+  "dependencies": {
+    "bufferutil": "^4.0.5",
+    "utf-8-validate": "^5.0.7"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "lib:test": "cd packages && cd lib && yarn test",
     "site:start": "cd packages && cd site && yarn start",
     "site:build": "cd packages && cd site && yarn build",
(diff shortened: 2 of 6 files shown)


xaviergonz:
looks great, thanks a lot!!

sisp:
I'm glad you like it. :slightly_smiling_face:

dodas:
love it, thanks!

PR #365, opened 2022-01-21
add prop transform "stringToBigIntTransform"

### LINKED ISSUE #363: Builtin prop transform for BigInt
[BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) is a useful alternative to `number` for representing (large) integers and has [good browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility) these days. However, there is no JSON representation for BigInt, so it needs to be encoded, e.g. as a string.

I think it would be nice to have a builtin prop transform for BigInt. It might not be overly difficult to implement a custom prop transform for it, but since BigInt is a native data type, I think it would be a nice feature of the library to have first-class support for it. I suggest to encode BigInt as a string of digits with a trailing `n`, e.g. `12345n`.

What do you think?

sisp:
Resolves #363.


sisp:
Code coverage is decreasing. I added no tests because the other prop transforms have no tests either. :laughing:

xaviergonz:
Actually there's a propTransform.test.ts for timestampToDateTransform

sisp:
You're right, I somehow missed that file.

I've added a test case there.

xaviergonz:
thanks!

PR #366, opened 2022-01-21
add runtime type-checker for bigint

sisp:


xaviergonz:
Hmm, the problem with a types.bigint is that it is not serializable to json, so it would result on unseralizable snapshots/patches. The other option would be to define `const types.bigint = typesRefinement(typesString, (n) => /* here check string is an integer number */, "bigint").withTransform(stringToBigIntTransform)` or similar, but that would be kind of new (there hasn't been a type with a built-in transform before, but I guess this would be as a good time as any to have it).

sisp:
Yes, you're right. Your observation is related to #369, something I stumbled over while working on those BigInt features.

xaviergonz:
That being said, if there's a `types.bigint` with a built-in transform probably there should be a `types.date("timestamp" | "isoString")` as well...

sisp:
Something like `types.<someType>.withTransform(...)` is along the lines of what I was vaguely trying to say in https://github.com/xaviergonz/mobx-keystone/issues/369#issuecomment-1019461236. Runtime types might ultimately not only check types but form a data schema including data encoding/decoding / serialization/deserialization + while at it also default values for optional (nested) fields.

sisp:
> That being said, if there's a types.bigint with a built-in transform probably there should be a types.date("timestamp" | "isoString") as well... (I'm not implying that should be part of this PR, just thinking out loud :) )

As an additional benefit, types with built-in transforms would be awesome when used in objects and nested fields:

```ts
tProp(types.object(() => ({
  int: types.bigint,
  date: types.date("timestamp"),
})))
```

Currently, it would be hard to do that with the `tProp(...).withTransform(...)` API.

xaviergonz:
I tried it, even had a whole branch dedicated to that idea, but in the end I got bitten by some of the same typing problems than MST (problems with recursive types mostly)...

sisp:
Hm, do you still have that branch and would you share it?

xaviergonz:
Would need to check if it's still alive (it was on the laptop that broke and not pushed anywhere)

PR #368, opened 2022-01-22
improve enum runtime type-checker typing

sisp:
More accurate and simpler typing of the enum runtime type-checker.


xaviergonz (approved):

xaviergonz:
thanks!

PR #371, opened 2022-01-24
remove internal function lazy and use lateVal instead

sisp:
This PR replaces uses of the internal utility function `lazy` by `lateVal` because `lateVal` is a more general version of `lazy`, so there's no need to have both.

Do you want to rename `lateVal` to `lazy` or keep it as `lateVal`?


sisp:
I think the tiny drop in coverage is irrelevant. I'm not even sure why that is.

xaviergonz:
Thanks! I'll just merge this and then rename it in another commit

PR #373, opened 2022-01-25
add action serialization for non-JSON primitive values

sisp:
This PR adds support for action serialization of non-JSON primitive values (`undefined`, `bigint`, and special `number` values `NaN`/`+Infinity`/`-Infinity`, but not `symbol` because I think there's no general way to serialize them) including tests.

There are likely several ways to implement it (single vs. multiple (de)serializer(s), encoding of those primitive values etc.). This is my suggestion, but feel free to request a different approach. :slightly_smiling_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 - Added property transform `stringToBigIntTransform`.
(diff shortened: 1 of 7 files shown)


xaviergonz:
thanks!

PR #376, opened 2022-01-28
use enum instead of string literals for action tracking middleware state

sisp:
Certainly a matter of taste/style, but I think enums are preferable to string literals for representing state. Feel free to prefer keeping it as it is now if you disagree. :slightly_smiling_face:

### DIFF
--- a/packages/lib/src/actionMiddlewares/actionTrackingMiddleware.ts
+++ b/packages/lib/src/actionMiddlewares/actionTrackingMiddleware.ts
@@ -141,12 +141,12 @@ export function actionTrackingMiddleware(
 
   const dataSymbol = Symbol("actionTrackingMiddlewareData")
   enum State {
-    Idle,
-    Started,
-    RealResumed,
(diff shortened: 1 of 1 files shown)


packages/lib/src/actionMiddlewares/actionTrackingMiddleware.ts
```diff
@@ -140,9 +140,17 @@ export function actionTrackingMiddleware(
...
 
   const dataSymbol = Symbol("actionTrackingMiddlewareData")
+  enum State {
+    Idle,
+    Started,
+    RealResumed,
+    FakeResumed,
+    Suspended,
+    Finished,
+  }
```
xaviergonz, line 150:
```suggestion
  enum State {
    Idle = "idle",
    Started = "started",
    RealResumed = "realResumed",
    FakeResumed = "fakeResumed",
    Suspended = "suspended",
    Finished = "finished",
  }
```

PR #378, opened 2022-01-29
improve [Simple]ActionContext data property typing

sisp:
Are you aware of any situation in which `data` should not be an object with `symbol` keys? If not, I think the typing should enforce it.


xaviergonz:
good idea, I don't see any reason not to

PR #383, opened 2022-02-03
add support for computed trees

### LINKED ISSUE #327: Manage computed model views as tree nodes
I have a situation in which I need to derive one model from another (i.e. create a model view, somewhat related to [`createTransformer`](https://www.mobxjs.com/refguide/create-transformer.html)) and would like to have access to all contexts and tree navigation functions like `getParent` from the view as well as refer to it via a root ref.

Let's make this more specific:

```ts
enum DataType {
  Int = "int",
  Float = "float",
}

@model("app/Tensor")
class Tensor extends Model({
  id: idProp,
  dtype: prop<DataType>().withSetter(),
  shape: prop<number[]>().withSetter(),
}) {
  @computed
  get ndim(): number {
    return this.shape.length
  }
  
  // ...
}

const dataFrameContext = createContext<DataFrame>()

// A `Block` is an extension (because it has header names) and subset (because
// it has only 2 dimensions) of a `Tensor`. I need the `Tensor` because I
// operate on `Tensor` instances elsewhere, but I need the `Block` to configure
// the `DataFrame`.
@model("app/Block")
class Block extends Model({
  id: idProp,
  columns: prop<string[]>().withSetter(), // header names
  dtype: prop<DataType>().withSetter(),
}) {
  // Derive the `Tensor` from the `Block`.
  @computed
  get tensor(): Tensor {
    return new Tensor({
      id: `${this.id}.tensor`, // to make the tensor view ID deterministic
      dtype: this.dtype,
      shape: [dataFrameContext.get(this)?.rows ?? 0, this.columns.length],
    })
  }

  // ...
}

@model("app/DataFrame")
class DataFrame extends Model({
  id: idProp,
  blocks: prop<Block[]>().withSetter(),
  rows: prop<number>().withSetter(),
}) {
  onInit(): void {
    dataFrameContext.set(this, this)
  }

  // ...
}
```

And now I'll instantiate a `DataFrame` and, e.g., would like to access the parent of the tensor view:

```ts
const df = new DataFrame({
  blocks: [
    new Block({ columns: 1, dtype: DataType.Int }),
    new Block({ columns: 10, dtype: DataType.Float }),
  ],
  rows: 100,
})

getParent(df.blocks[0].tensor) // is `undefined` but I'd like it to be the `Block` instance
```

In order to achieve my desired behavior, I've been wondering whether it makes sense to add a new decorator called, e.g., `@modelView` which is `@computed` + attaching/detaching of the returned model. A `@modelView`-decorated getter must always return a model instance which must not already be attached anywhere else.

What do you think about this idea? And does my example make any sense, or does my modeling approach look wrong?

sisp:
This PR adds support for computed trees. What do you think about this implementation?

In the end, I went for the monolithic `@computedTree` decorator which internally calls MobX' `@computed` decorator instead of going for a modular decorator like `@tree` that requires a getter to be already decorated with `@computed`. I think it aligns better with the rest of the library, e.g. `@modelAction` also turns a method into a MobX action internally instead of requiring MobX' `@action` decorator explicitly. What do you think?

All tree traversal functions except `resolvePath` work with computed trees (see the tests). `resolvePath` would need to be adapted a bit because it automatically navigates to the data object (`<model>.$`) of a model instance and, thus, cannot traverse to a computed property. It might be possible to fall back to checking the model instance for a property in case it's not found in the data object. Not sure if that disagrees with the intention of `resolvePath`.  The corresponding test is currently skipped because it fails but to make sure it's a known problem.

Closes #327.


packages/lib/src/computedTree/computedTree.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param target Prototype of the class.
+ * @param propertyKey Name of the member.
+ * @param descriptor Property descriptor for the member.
+ */
+export function computedTree(
+  target: any,
+  propertyKey: string,
+  descriptor: PropertyDescriptor
+): void {
```
xaviergonz, line 34: are you sure this decorator works with both babel and tsc? (I think they used to generate different descriptor code)

packages/lib/test/computedTree/computedTree.test.ts
```diff
@@ -0,0 +1,518 @@
...
+  expect(() => {
+    // @ts-ignore
+    // eslint-disable-next-line
+    class X extends Model({}) {
+      @computedTree
+      set value(_value: any) {}
+    }
+  }).toThrow("@computedTree requires a 'get' accessor")
+})
+
```
xaviergonz, line 154: (optional) cannot decorate a method / cannot decorate a field

packages/lib/test/computedTree/computedTree.test.ts
```diff
@@ -0,0 +1,518 @@
...
+    expect(isParentOfChild(r, r.model)).toBeTruthy()
+  })
+
+  // fails because `resolvePath` skips ahead to `<model>.$` when it arrives at a model node
+  test.skip("computed tree supports resolvePath", () => {
+    const r = new R({})
+
+    expect(resolvePath(r, ["array"])).toEqual({ resolved: true, value: r.array })
+    expect(resolvePath(r, ["plainObject"])).toEqual({ resolved: true, value: r.plainObject })
+    expect(resolvePath(r, ["model"])).toEqual({ resolved: true, value: r.model })
```
xaviergonz, line 279: perhaps resolvePath should be changed to also check the original object (not the $ property) but only if nothing is found inside the $ property before "failing"?

packages/lib/test/computedTree/computedTree.test.ts
```diff
@@ -0,0 +1,518 @@
...
+    expect(children.has(r.plainObject)).toBeTruthy()
+    expect(children.has(r.model)).toBeTruthy()
+  })
+
+  test.each([WalkTreeMode.ChildrenFirst, WalkTreeMode.ParentFirst])(
+    "computed tree supports walkTree (mode=%s)",
+    (mode) => {
+      const r = new R({})
+
+      // does not traverse computed trees because they are not obesrved at this point
```
xaviergonz, line 344:
```suggestion
      // does not traverse computed trees because they are not observed at this point
```

packages/lib/test/computedTree/computedTree.test.ts
```diff
@@ -0,0 +1,518 @@
...
+          () => [r.array, r.plainObject, r.model],
+          () => {}
+        )
+      )
+
+      expect(walkTree(r, (node) => (isArray(node) ? node : undefined), mode)).toBe(r.array)
+      expect(walkTree(r, (node) => (isPlainObject(node) ? node : undefined), mode)).toBe(
+        r.plainObject
+      )
+      expect(walkTree(r, (node) => (node instanceof M ? node : undefined), mode)).toBe(r.model)
```
xaviergonz, line 361: hmm, I'm not sure if it is desirable for computed trees to appear in walkTree at all...

packages/lib/test/computedTree/computedTree.test.ts
```diff
@@ -0,0 +1,518 @@
...
+        () => [r.array, r.plainObject, r.model],
+        () => {}
+      )
+    )
+
+    const children = getChildrenObjects(r, { deep })
+    expect(children.size).toBe(3)
+    expect(children.has(r.array)).toBeTruthy()
+    expect(children.has(r.plainObject)).toBeTruthy()
+    expect(children.has(r.model)).toBeTruthy()
```
xaviergonz, line 336: I'm not sure if it is good that computed tree nodes appear in getChildrenObjects (not to mention having them magically appear/disappear when they are observed/unobserved is confusing)

xaviergonz: perhaps isComputedTreeNode could be used to filter them out?

xaviergonz: or at least they should be hidden under an optional parameter which is false (so they are hidden) by default

packages/lib/test/computedTree/computedTree.test.ts
xaviergonz: perhaps isComputedTreeNode could be used before deciding if the callback should be called?

xaviergonz:
Wouldn't this feature generate a mem leak every time the computed property is called while being unobserved?
E.g., if the getter generates a new model node and assigns itself as children of the original model then it will get added to the "children" list of such object, never to be released.

xaviergonz:
Ah, I see you use a property key when setting its parent. Still, it would be good to double check it is not the case :)

xaviergonz (changes requested):
left a few comments :)

packages/lib/test/computedTree/computedTree.test.ts
sisp: Are you suggesting to change the error message or add tests that check that the decorator cannot decorate a method or field?

packages/lib/test/computedTree/computedTree.test.ts
sisp: Hm, if `walkTree` didn't call the callback for computed tree nodes, then the `onAttachedToRootStore` hook wouldn't be called for those nodes either because [root store attachment/detachment uses `walkTree` internally](https://github.com/xaviergonz/mobx-keystone/blob/9c11ff21d64512ca3bc3ceb5a953eab2cc4a51f3/packages/lib/src/rootStore/attachDetach.ts#L22-L43). We _could_ introduce an option like `includeComputedTreeNodes: boolean` and exclude computed tree nodes by default for backwards compatibility. In the root store attachment/detachment code, it would need to be set to `true` though which (I think) introduces some coupling between the root store feature and the computed tree feature.

Also, my view on computed trees is that they are essentially equivalent to regular trees except they are immutable and, thus, action middlewares are useless because there are no action calls in computed trees. With that in mind, I think they should be treated no different than regular trees as much as possible, so e.g. tree traversal functions should behave the same for computed and regular trees.

But I'm more than happy to be convinced otherwise. :slightly_smiling_face:

packages/lib/test/computedTree/computedTree.test.ts
sisp: I've changed `resolvePath` accordingly (d09dc8b406568e888ffdeccd1856d7bf8cbe8f2c).

packages/lib/src/computedTree/computedTree.ts
sisp: Good point, I hadn't considered that. I'll check and get back to you.

packages/lib/test/computedTree/computedTree.test.ts
sisp: This is somewhat related to [your comment about `walkTree`](https://github.com/xaviergonz/mobx-keystone/pull/383#discussion_r799012501). Whatever we decide there, I think `walkTree` and `getChildrenObjects` (and `findChildren`) should be consistent.

> [...] (not to mention having them magically appear/disappear when they are observed/unobserved is confusing)

I'm not sure whether I find it confusing or not. Isn't it similar to having a node attached to (= observed) or detached from (i.e. set to `undefined`; = unobserved) a model prop?

As mentioned in my `walkTree` comment:

> Also, my view on computed trees is that they are essentially equivalent to regular trees except they are immutable and, thus, action middlewares are useless because there are no action calls in computed trees. With that in mind, I think they should be treated no different than regular trees as much as possible, so e.g. tree traversal functions should behave the same for computed and regular trees.

From that perspective, I think it makes sense to me the way it is now. But I'm happy to be convinced otherwise. :slightly_smiling_face:

That said, would it make sense to set [`keepAlive: true`](https://mobx.js.org/computeds.html#keepalive) on the computed tree? This way, the computed property would always be observed and the behavior would be more consistent with regular tree nodes.

sisp: Oh, and if `getChildrenObjects` filtered out computed tree nodes, then it would be impossible to call it on a computed tree node as the subtree root, too, so it would be entirely useless in computed trees.

packages/lib/test/computedTree/computedTree.test.ts
sisp: Oh, and if `walkTree` filtered out computed tree nodes, then it would be impossible to call it on a computed tree node as the subtree root, too, so it would be entirely useless in computed trees.

packages/lib/src/computedTree/computedTree.ts
sisp: I've checked how decorators are implemented by Babel, mainly by reverse-engineering it via the Babel REPL. TS + decorators (legacy) + loose (`true` & `false`) has the same decorator signature as [TS accessor decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#accessor-decorators) (see [REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=true&code_lz=GYVwdgxgLglg9mABAEwKYTgJwIZSwCgEpEBvAKEUQzAGc4AbVAOnrgHN9tM2QBbVMFBqEyAXzJkI9bDRqIAgqQqIAAmgw48mZW1RREANyJLKASEx6QmJAFZl48ZIR1GLdvjCoA7gqJMDhEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Ctypescript&prettier=true&targets=&version=7.17.1&externalPlugins=&assumptions=%7B%7D)). So I'd say the current implementation should work for both TSC and Babel, but I don't have much experience with the quirks of decorators yet.

Do you want to support other decorator proposals than "legacy"?

It would be nice to test this as part of CI/CD though. :thinking:

xaviergonz: Those should be good enough for now :)

packages/lib/test/computedTree/computedTree.test.ts
xaviergonz: The last

packages/lib/test/computedTree/computedTree.test.ts
xaviergonz: > We could introduce an option like includeComputedTreeNodes: boolean and exclude computed tree nodes by default for backwards compatibility

I think that would make sense, or maybe even an enum: 'nonComputed' | 'computed' | 'all'

packages/lib/test/computedTree/computedTree.test.ts
xaviergonz: Maybe it could use an option to get those, like walkTree?

xaviergonz: > That said, would it make sense to set [keepAlive: true](https://mobx.js.org/computeds.html#keepalive) on the computed tree? This way, the computed property would always be observed and the behavior would be more consistent with regular tree nodes.

I think keepAlive would never ever release it (not 100% sure though), but yeah, perhaps there could be a "custom keep alive", a WeakMap<nonComputedObject, Set<ComputedObject>> that would be auto populated when the nonComputedObject is constructed (right before onInit?). If that were the case I'd have no problem keeping those nodes in walkTree/getChildrenObjects (and would be even preferable so there's no filtering overhead)

xaviergonz: https://medium.com/terria/when-and-why-does-mobxs-keepalive-cause-a-memory-leak-8c29feb9ff55#:~:text=Why%20use%20keepAlive%20%3F,computed%20value%20is%20automatically%20cached.&text=When%20one%20of%20the%20observables,and%20re%2Drun%20the%20derivation.

packages/lib/test/computedTree/computedTree.test.ts
sisp: I've added a test that checks decorating a method. When decorating a field, TS complains already:

> Unable to resolve signature of property decorator when called as an expression. ts(1240)

Would you like a test for that case anyway? For fields, the descriptor is `undefined`, so I'd probably need to make it optional check this case, although when applied to an accessor the descriptor would never be `undefined`, i.e. I'd have to change the typing for a case that TS would disallow anyway.

packages/lib/test/computedTree/computedTree.test.ts
sisp: Thanks for digging into the `keepAlive` topic! I agree, it looks like `keepAlive: true` should be fine. From what I could tell after settings this option, it is not immediate, so I added another late initialization hook just before `onInit`/`onLazyInit` and registered a callback that access the property ones. See ce5e5d1e4f96371eb8d6d5b761a68f51286623d7.

It seems to work fine after a couple of changes because eager evaluation caused infinite recursion with the previous test setup. The only thing I'm not able to figure out right now is why the `onChildAttachedTo` test fails (it's being skipped now). The callback is not called after `r.setValue(10)`, but `r.model.value` is `10` then. Perhaps this change is not considered a detachment of the old model instance and attachment of the new model instance?

xaviergonz: It happened because the old value was never getting detached first. I worked around it by using a cache of "tweaked values" and untweaking old values / tweaking new values when the inner computed value changed

xaviergonz: https://github.com/xaviergonz/mobx-keystone/pull/383/commits/d58cda24bcdfbf0be6a79146652a4e55c4e712fe

packages/lib/test/computedTree/computedTree.test.ts
xaviergonz: nah, good enough

xaviergonz (approved):
lgtm, let me know when it should be merged

xaviergonz:
btw, do you think the first run should actually be done before onInit or after onInit?

packages/lib/test/computedTree/computedTree.test.ts
sisp: Very nice! Thank you very much for fixing it! :pray:

I only made a minor change: I simplified `getComputedTreeNodeInfo`  by using the `getOrCreate` utility function (see 07a2176db75f06168382c43020a1ec6d92a4d41d).

sisp:
> btw, do you think the first run should actually be done before onInit or after onInit? (thinking of a computed node that might actually require some kind of initialization?)

Hm, it looks to me like there is no explicit "first run" anymore since d58cda24bcdfbf0be6a79146652a4e55c4e712fe. The way it works now, I think the late initialization block can't be run _after_ `onInit` because the modified getter requires `getComputedTreeNodeInfo` to be populated with the object corresponding to that instance and property, and this implies accessing the property once. Perhaps this block could be run the first time the property is accessed instead (a lazy initialization).

That said, I think performing the first run _after_ `onInit` would be inconsistent with regular trees via model props because tree traversal functions like `getChildrenObjects` work for regular trees in `onInit` but wouldn't work for computed trees before the first run. Also, at least right now, I can't think of a case where the first run before `onInit` would be a problem. Unless you've found a problematic case, I'd leave it as it is for now.

> lgtm, let me know when it should be merged

I've made a few more minor commits including docs improvements, mainly added a note about the differences between a computed tree property and a MobX computed property. I think it's ready to be merged now.

sisp:
Thank you very much for all your support in ensuring the quality of this feature both in terms of behavior and implementation! :pray:  I'm genuinely amazed how the initial draft implementation has evolved over the course of your reviews and our discussions, the result being a feature that feels sound and consistent with the rest of the library. :ok_hand:

dodas:
wohooo, what a nice sunday gift! I am going to play with this now.
thank you both! 🚀

xaviergonz:
thanks to you for implementing it!!

PR #389, opened 2022-02-07
run tests against all supported MobX versions

sisp:
This PR adds support for running the test suite against all officially supported MobX versions (according to [`packages/lib/package.json`](https://github.com/xaviergonz/mobx-keystone/blob/410068bb1b0167eef5b84d7d0737eb2c28c96809/packages/lib/package.json#L66)).

It leverages Yarn package aliases to install multiple versions of the same package simultaneously and a Jest mock using [`jest.requireActual(...)`](https://jestjs.io/docs/jest-object#jestrequireactualmodulename) for transparently directing the import to the desired package version.

Right now, the test suite only passes for MobX v6 which reveals that v5 and v4 are actually not 100% supported at the moment. With MobX v5, only a few tests fail, so I imagine it might be possible to fix those problems. With MobX v4, a large number of the tests fail and the test suite even gets stuck at some point. For this reason, I've not added MobX v4 to the CI matrix for now and only wrote a TODO comment to remember this entry is missing. You can try it locally like this:

```bash
$ MOBX_VERSION=4 yarn lib:test
```

What do you think about this situation? Should we aim for supporting MobX v4/v5/v6 or only v5/v6, or even only v6?


sisp:
Hm, I'm not sure why the job for MobX v6 was canceled. :thinking: Locally, it passes for me.

xaviergonz:
for it to work it would need to also change tsconfig for each mobx version (particularly use define for class fields needs to be true for mobx 6 and false for mobx 4 (maybe also 5)?

xaviergonz:
about the job cancel, maybe the jobs should include the mobx version in the cache keys

sisp:
I did some digging and it's actually related to the [`fail-fast`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) setting in the `strategy` section. The cache shouldn't be affected because Yarn installs all versions of MobX simultaneously.

sisp:
`useDefineForClassFields: false` fixes all but one test for MobX v5. MobX v4 still fails heavily. :thinking:

xaviergonz:
I just pushed some changes to make it compatible with v5. I think we can just skip mobx 4 for now (I'm not even sure if it is compatible to be honest) but leave the stuff in place just in case somebody might want to give it a try?

xaviergonz:
btw, mobx 5 doesn't change useDefineForClassFields: true, mobx 4 is the one that requires false

sisp:
I just committed a simplification of the dynamic MobX module switching for the different versions using Jest's `moduleNameMapper` config instead of mocking.

xaviergonz:
I'll merge it now, but I might eventually fix v4 :) thanks!!

PR #391, opened 2022-02-11
run tests also with Babel

sisp:
This PR adds support for running the test suite with Babel in addition to `tsc` (via `ts-jest`), especially regarding decorator compatibility with Babel's transpilation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,15 @@ jobs:
       - run: yarn lib:build && yarn lib:build-docs && yarn site:build
 
   lib-test:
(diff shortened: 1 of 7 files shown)


sisp:
One test is failing, but I don't understand why ... :thinking:

xaviergonz:
Because of this babel decorators plugin bug: https://github.com/babel/babel/issues/11131
I'll push a fix in a sec

xaviergonz:
Thanks!!

sisp:
Ah, again that bug ... Thanks for the fix! 🙂

PR #399, opened 2022-02-19
use @internal tag instead of @ignore

sisp:
Some arguments in favor of this change:

- TypeDoc also hides `@internal` tagged code from the docs when [`stripInternal` is set to `true` in `tsconfig.json`](http://typedoc.org/guides/options/#excludeinternal) (which it is), so `@ignore` + `@internal` is redundant.
- `@internal` strips code from the public API of the library and typically undocumented (`@ignore` tagged) code isn't intended to be part of the public API, so it's safer to use `@internal`. There's one exception: the public ("private") methods in the `UndoStore` model (e.g. `_clearUndo()`, ...) should be `@ignore` tagged in my opinion to keep them in the emitted `.d.ts` file.

### DIFF
--- a/packages/lib/src/dataModel/BaseDataModel.ts
+++ b/packages/lib/src/dataModel/BaseDataModel.ts
@@ -163,7 +163,7 @@ export abstract class BaseDataModel<TProps extends ModelProps> {
 setBaseDataModel(BaseDataModel)
 
 /**
- * @ignore
+ * @internal
  */
 export type BaseDataModelKeys = keyof AnyDataModel | "onLazyInit"
 
--- a/packages/lib/src/dataModel/BaseDataModel.ts
+++ b/packages/lib/src/dataModel/BaseDataModel.ts
@@ -163,7 +163,7 @@ export abstract class BaseDataModel<TProps extends ModelProps> {
 setBaseDataModel(BaseDataModel)
 
 /**
- * @internal
+ * @ignore
  */
 export type BaseDataModelKeys = keyof AnyDataModel | "onLazyInit"
 
--- a/packages/lib/src/frozen/Frozen.ts
+++ b/packages/lib/src/frozen/Frozen.ts
@@ -16,7 +16,7 @@ export enum FrozenCheckMode {
 }
 
 /**
- * @internal
+ * @ignore
  */
 export const frozenKey = "$frozen"
 
--- a/packages/lib/src/frozen/Frozen.ts
+++ b/packages/lib/src/frozen/Frozen.ts
@@ -16,7 +16,7 @@ export enum FrozenCheckMode {
 }
 
 /**
- * @ignore
+ * @internal
  */
 export const frozenKey = "$frozen"
 
@@ -118,7 +118,6 @@ function checkDataIsSerializableAndFreeze(data: any) {
 }
 
 /**
- * @ignore
  * @internal
  *
  * Checks if an snapshot is an snapshot for a frozen data.
--- a/packages/lib/src/model/BaseModel.ts
+++ b/packages/lib/src/model/BaseModel.ts
@@ -25,7 +25,7 @@ import { assertIsModelClass } from "./utils"
 import { setBaseModel } from "./_BaseModel"
 
 /**
- * @ignore
+ * @internal
  */
 export const modelIdPropertyNameSymbol = Symbol()
 
@@ -165,7 +165,7 @@ export abstract class BaseModel<
 setBaseModel(BaseModel)
 
 /**
- * @ignore
+ * @internal
  */
 export type BaseModelKeys = keyof AnyModel | "onInit" | "onAttachedToRootStore"
 
--- a/packages/lib/src/model/BaseModel.ts
+++ b/packages/lib/src/model/BaseModel.ts
@@ -27,7 +27,7 @@ import { setBaseModel } from "./_BaseModel"
 /**
  * @ignore
  */
-export const modelIdPropertyNameSymbol = Symbol()
+export const modelIdPropertyNameSymbol = Symbol("modelIdPropertyName")
 
 /**
  * Base abstract class for models. Use `Model` instead when extending.
--- a/packages/lib/src/model/BaseModel.ts
+++ b/packages/lib/src/model/BaseModel.ts
@@ -25,7 +25,7 @@ import { assertIsModelClass } from "./utils"
 import { setBaseModel } from "./_BaseModel"
 
 /**
- * @internal
+ * @ignore
  */
 export const modelIdPropertyNameSymbol = Symbol()
 
@@ -165,7 +165,7 @@ export abstract class BaseModel<
 setBaseModel(BaseModel)
 
 /**
- * @internal
+ * @ignore
  */
 export type BaseModelKeys = keyof AnyModel | "onInit" | "onAttachedToRootStore"
 
--- a/packages/lib/src/modelShared/BaseModelShared.ts
+++ b/packages/lib/src/modelShared/BaseModelShared.ts
@@ -13,17 +13,17 @@ import type {
 /**
  * @ignore
  */
-export const propsTypeSymbol = Symbol()
+export const propsTypeSymbol = Symbol("propsType")
 
 /**
  * @ignore
  */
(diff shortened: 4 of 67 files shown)


packages/lib/src/dataModel/BaseDataModel.ts
```diff
@@ -163,7 +163,7 @@ export abstract class BaseDataModel<TProps extends ModelProps> {
 setBaseDataModel(BaseDataModel)
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 166: this cannot be internal since it is used as part of the definition of the "public" interface _DataModel

packages/lib/src/frozen/Frozen.ts
```diff
@@ -16,7 +16,7 @@ export enum FrozenCheckMode {
 }
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 19: it is used by public interfaces

xaviergonz: SnapshotOutOfFrozen etc

packages/lib/src/model/BaseModel.ts
```diff
@@ -25,7 +25,7 @@ import { assertIsModelClass } from "./utils"
 import { setBaseModel } from "./_BaseModel"
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 28: used by public interfaces

packages/lib/src/model/BaseModel.ts
```diff
@@ -165,7 +165,7 @@ export abstract class BaseModel<
 setBaseModel(BaseModel)
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 168: used by public interfaces

packages/lib/src/modelShared/BaseModelShared.ts
```diff
@@ -11,22 +11,21 @@ import type {
...
- * @ignore
+ * @internal
  */
 export const fromSnapshotOverrideTypeSymbol = Symbol()
 
 /**
- * @ignore
+ * @internal
  */
 export const toSnapshotOverrideTypeSymbol = Symbol()
```
xaviergonz, line 26: used by public interfaces

packages/lib/src/modelShared/prop.ts
```diff
@@ -4,7 +4,7 @@ import { getOrCreate } from "../utils/mapUtils"
 import type { Flatten, IsNeverType, IsOptionalValue } from "../utils/types"
 
 /**
- * @ignore
+ * @internal
  */
 export const noDefaultValue = Symbol("noDefaultValue")
```
xaviergonz, line 9: used by public interfaces

packages/lib/src/modelShared/prop.ts
```diff
@@ -260,7 +260,7 @@ export const idProp = {
 } as any as ModelIdProp
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 263: used by public interfaces

packages/lib/src/ref/Ref.ts
```diff
@@ -60,7 +60,7 @@ export abstract class Ref<T extends object> extends Model({
 }
 
 /**
- * @ignore
+ * @internal
  */
 export declare const customRefTypeSymbol: unique symbol
```
xaviergonz, line 65: used by public interfaces

packages/lib/src/types/primitiveBased/enum.ts
```diff
@@ -4,7 +4,7 @@ import { typesOr } from "../utility/or"
 import { typesLiteral } from "./primitives"
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 7: used by public api

packages/lib/src/types/primitiveBased/enum.ts
```diff
@@ -27,7 +27,7 @@ function enumValues(e: EnumLike): (string | number)[] {
 }
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 30: used by public api

packages/lib/src/types/schemas.ts
```diff
@@ -8,10 +8,10 @@ import type { IsOptionalValue } from "../utils/types"
 // infer is there just to cache type generation
 
 export interface Type<Name, Data> {
-  /** @ignore */
+  /** @internal */
```
xaviergonz, line 11: this needs to be in the d.ts or else type inference will be wrong

packages/lib/src/types/schemas.ts
```diff
@@ -8,10 +8,10 @@ import type { IsOptionalValue } from "../utils/types"
...
   $$type: Name
 
-  /** @ignore */
+  /** @internal */
```
xaviergonz, line 14: this needs to be in the d.ts or else type inference will be wrong

packages/lib/src/types/schemas.ts
```diff
@@ -28,7 +28,7 @@ export interface ArrayType<S> // e.g. S = someType[] or [someType, someType]
   > {}
 
 export interface ObjectOfTypes {
-  /** @ignore */
+  /** @internal */
```
xaviergonz, line 31: this needs to be in the d.ts or else type inference will be wrong

packages/lib/src/types/schemas.ts
```diff
@@ -106,5 +106,5 @@ export type TypeToData<S> = S extends ObjectTypeFunction
   : // anything else
     never
 
-/** @ignore */
+/** @internal */
```
xaviergonz, line 109: part of the public api

packages/lib/src/utils/types.ts
```diff
@@ -1,19 +1,19 @@
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 2: used by public api

packages/lib/src/utils/types.ts
```diff
@@ -1,19 +1,19 @@
...
  *
  * A primitive value.
  */
 export type PrimitiveValue = undefined | null | boolean | number | string | bigint
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 9: used by public api

packages/lib/src/utils/types.ts
```diff
@@ -1,19 +1,19 @@
...
  *
  * A JSON-compatible primitive value.
  */
 export type JSONPrimitiveValue = null | boolean | number | string
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 16: used by public api

packages/lib/src/utils/types.ts
```diff
@@ -38,11 +38,11 @@ export type IsOptionalValue<C, TV, FV> = IsNeverType<Extract<C, undefined>, FV,
 // type _F = IsOptionalValue<unknown, true, false> // false, but we don't care
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 41: used by public api

packages/lib/src/utils/types.ts
```diff
@@ -38,11 +38,11 @@ export type IsOptionalValue<C, TV, FV> = IsNeverType<Extract<C, undefined>, FV,
...
  */
 export type IsNeverType<T, IfNever, IfNotNever> = [T] extends [never] ? IfNever : IfNotNever
 
 /**
- * @ignore
+ * @internal
```
xaviergonz, line 46: used by public api

xaviergonz:
Thanks! I reverted some types being marked as internal and put them back as ignored since they are indirectly being used by public types.

sisp:
Ouch, thanks for catching those errors of mine! It would be great if there was a way to test the declarations files for consistency to prevent such errors from happening. Would it be enough to somehow check the generated `.d.ts` files? Or would it require running additional typing tests against those declarations? 🤔

xaviergonz:
I think it would be enough to type check all generated d.ts files

Maybe some .ts file that just imports the main index.d.ts? and does a console log and compile that? (with skipLibCheck or whatever the name set to false)

Issue #405, opened 2022-02-20
TS typing error with generic factory function for generic model instance


sisp:
I've discovered another typing issue related to generic models when using a generic factory function to create a new model instance. This is a test case that reproduces the problem:

```ts
test("generic model instance factory", () => {
  // @model("Parent")
  class Parent<T> extends Model(<T>() => ({
    a: prop<T>(),
  }))<T> {}

  // @model("Child")
  class Child<T> extends ExtendedModel(<T>() => ({
    baseModel: modelClass<Parent<T>>(Parent),
    props: {
      b: prop<T>(),
    },
  }))<T> {}

  // @ts-ignore
  // eslint-disable-next-line
  function createParent<T>(a: T) {
    const parent = new Parent({ a })
    assert(_ as typeof parent, _ as Parent<T>) // ERROR
  }

  // @ts-ignore
  // eslint-disable-next-line
  function createChild<T>(a: T, b: T) {
    const child = new Child({ a, b })
    assert(_ as typeof child, _ as Child<T>) // ERROR
  }
})
```

I've tried to fix it but have not been successful so far.

xaviergonz:
Fixed in this commit https://github.com/xaviergonz/mobx-keystone/commit/8f8d22b95829b3363320d12e2262d13a56474985

xaviergonz:
v0.66.1

sisp:
Awesome! :tada: Thank you very much for the prompt fix!

sisp:
I'm afraid it's not fully resolved. When you add an ID prop, a typing error occurs again:

```diff
test("generic model instance factory", () => {
   @model("generic model instance factory/Parent")
   class Parent<T> extends Model(<T>() => ({
+    id: idProp,
     a: prop<T>(),
   }))<T> {}
```

sisp:
Actually, it's not about the ID prop, but the problem is related to any optional prop, e.g.:

```diff
test("generic model instance factory", () => {
   @model("generic model instance factory/Parent")
   class Parent<T> extends Model(<T>() => ({
+    x: prop<number | undefined>(),
     a: prop<T>(),
   }))<T> {}
```

xaviergonz:
I think T being there is messing up the detection of optional properties, but as a workaround if you pass `x: undefined` explicitly it works

sisp:
Hm, yes, but I can't pass `undefined` for `id`. :thinking:

sisp:
I think I found a fix - see #407.

[...]

PR #407, opened 2022-02-21
fix issue related to constructor generics with optional props

sisp:
Fixes https://github.com/xaviergonz/mobx-keystone/issues/405#issuecomment-1047000482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+- Fixed typing issue related to generics with optional props.
+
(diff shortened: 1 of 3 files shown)


xaviergonz:
That's a nice trick! I didn't think of that 💯 
I've also used the same trick for the other omits :)

sisp:
Thanks! :blush:

> I've also used the same trick for the other omits :)

I had done that, too, but it wasn't needed to fix the constructor generics issue. Probably doesn't hurt though. :wink:

PR #409, opened 2022-02-22
use "in" operator instead of comparing with "undefined"

sisp:
I think the `in` operator is more readable and it's more type-safe (no `any` assertions).


xaviergonz:
Thanks! I reverted two checks were I think it is better to check for undefined instead (just in case the prop is there, but set to undefined for some reason)

sisp:
Alright :slightly_smiling_face:

PR #410, opened 2022-02-22
remove unneeded "any" assertions

sisp:
TS is happy also without those `any` assertions.


xaviergonz:
guess ts or the typings got better :) thanks!

PR #411, opened 2022-02-23
run tests also with SWC

sisp:
This PR adds support for running the test suite with [SWC](https://swc.rs/). SWC seems to be (becoming) quite popular (e.g. [Next.js introduced SWC in v12.0](https://nextjs.org/blog/next-12#faster-builds-and-fast-refresh-with-rust-compiler)), and it's written in Rust :sunglasses:. Especially since Next.js uses SWC by default now, I think it's worth making sure this library is compatible.

There was only one minor problem related to enums in order to make the test suite pass with SWC. The following check was failing:

https://github.com/xaviergonz/mobx-keystone/blob/d09658615af53c0e47c4d83d3daf87772f391b51/packages/lib/test/types/typeChecking.test.ts#L984

When I debugged it, it turned out that SWC compiles the enum

```ts
enum A {
  X1,
  X15 = "x15",
  X2 = 6,
}
```

to

```ts
{
  "0": "X1",
  "6": 6,
  X1: 0,
  X15: "x15",
  x15: "x15",
  X2: 6,
}
```

while `tsc` and Babel compile it to:

```ts
{
  "0": "X1",
  "6": "X2",
  X1: 0,
  X15: "x15",
  X2: 6,
}
```

So, I've adapted the `enumValues` function to handle this weird behavior of SWC, too.

Surprisingly though, [the compiled enum in SWC Playground looks fine](https://play.swc.rs/?version=1.2.144&code=H4sIAAAAAAAAA0vNK81VcFSo5lIAgghDHShtqmCroFRhaKoEFTAC8s10uGoB7sSgdi8AAAA%3D&config=H4sIAAAAAAAAA0WNSwrDMAxE76J1ts0id%2BghjKsEF%2F%2FQKBBjfPfawSU7aebpqdIXlrZK2QhYxoQS1Vy0kZbMsOKy0kKKHu3Gg1tfjBysHWG8eudTAs92oeCi28sw2RSyMPBUJh7%2BT7YuCulzjqDez27hSu1xzDuH9wRVTm4%2FOkwQUrUAAAA%3D). :thinking: Can you think of an explanation?


xaviergonz:
looks like a swc bug to me, I just reported it
https://github.com/swc-project/swc/issues/3711

btw, it happened to me in the SWC playground

packages/lib/src/types/primitiveBased/enum.ts
```diff
@@ -19,7 +19,8 @@ function enumValues(e: EnumLike): (string | number)[] {
     // we have to do this since TS does something weird
     // to number values
     // Hi = 0 -> { Hi: 0, 0: "Hi" }
-    if (typeof v !== "string" || e[v] !== +k) {
+    // and SWC does something weird, too
+    if (!vals.includes(v) && ((typeof v !== "string" && v !== +k) || e[v] !== +k)) {
```
xaviergonz, line 23: maybe this function should be unit tested with string enums, number enums and mixed number/string enums?

sisp: Agreed, see [583aa10](https://github.com/xaviergonz/mobx-keystone/pull/411/commits/583aa10c98370ccebeb7fc33032680fc8ae17562).

sisp:
> btw, it happened to me in the SWC playground

You're right, I didn't compare the results in the different playgrounds and didn't look closely enough. :facepalm:

xaviergonz:
Apparently the SWC issue got a fix merged and will be fixed in the next release

xaviergonz (approved):
feel free to merge when ready

sisp:
What do you think, should I revert the change to the `enumValues` function or keep it in case somebody still uses an SWC version before 1.2.145?

sisp:
> feel free to merge when ready

I don't think I have the necessary permissions.

Unless you would like the change to `enumValues` reverted, feel free to merge it any time.

xaviergonz:
since it doesn't break any cases let's leave it

PR #414, opened 2022-02-26
Pin benchmark chalk dependency to pre-ESM version

airhorns:
The benchmark was failing to compile because chalk v5.0.0 is ESM only which the build system is not yet ok with. This reverts the dependency bump back to a non ESM version which makes the benchmarks run again.


xaviergonz:
> Should we run the benchmark also during CI to check that it works?

Perhaps just check it builds is enough?

xaviergonz:
Thanks!

PR #418, opened 2022-02-28
add CI job that tests building the benchmark

sisp:
See https://github.com/xaviergonz/mobx-keystone/pull/414#issuecomment-1053742485.


xaviergonz:
Thanks!

PR #421, opened 2022-03-03
update compiler configs and docs

sisp:
It seems using different compiler settings for MobX 4 and 5/6 isn't necessary. Do you think this is generally valid? At least the test suite passes.


xaviergonz:
I don't know, I think those config were taken directly from mobx docs, so while they might work with keystone they might not work if you use decorators for mobx itself

sisp:
Okay, I'll investigate a bit further.

xaviergonz:
outdated

PR #425, opened 2022-03-06
improve badges style and README design

sisp:
An attempt at tuning the "marketing" appearance. :wink: Much inspired by https://github.com/vercel/next.js.

What do you think?

PR #426, opened 2022-03-07
fix badge aria label

sisp:
A small fix of my negligence in #425.


xaviergonz:
I think you left the link removal from the other pr on this PR too

sisp:
Ooops, you're right, sorry about that. I've reverted the deletion.

PR #427, opened 2022-03-07
remove redundant docs link

sisp:
Totally subjective, but I think the reference to the docs is a bit redundant. There's this one, there's one at the bottom of the README, and there's the link in the GitHub project's "About" section in the top-right of the landing page. I think the README looks a bit cleaner without this one. But I don't insist on its removal if you feel differently. :wink:


sisp:
> people don't read that much these days :sweat_smile:

I wish I could disagree. :rofl:

I've reverted the deletion and deleted the one at the bottom of the README.

PR #428, opened 2022-03-08
use @typeparam tag consistently

sisp:
Both `@typeparam` and `@template` are fine, but I think it's good to stick with one and be consistent. In almost all cases, `@typeparam` was used, so that's what I went for.


xaviergonz:
typeparam is the one typedocs supports, thanks!

sisp:
Actually, [it supports both](http://typedoc.org/guides/doccomments/#%40typeparam-%3Cparam-name%3E-or-%40template-%3Cparam-name%3E), but I think using one of them consistently is better than mixing them.

xaviergonz:
ah good to know! it didn't when I used to use it long ago

Issue #433, opened 2022-03-12
Add icon to simpleicons.org


sisp:
With your permission @xaviergonz I'd like to add mobx-keystone's icon to https://simpleicons.org. It's a large collection of brand icons and, e.g., shields.io uses it for badge icons. It already includes MobX' and MST's icons, so I think mobx-keystone shouldn't be missing.

Do you have an SVG version of the icon and would you mind sharing it?

xaviergonz:
That'd be cool, but I'm afraid I don't have a SVG version of the logo :-/

sisp:
Ah, no worries, I'll derive an SVG version then.

sisp:
Just noticed that in their contribution guidelines they request a GitHub project to have 5k stars or more. :confused: MobX and MST meet this requirement. It might be difficult to convince them that mobx-keystone is actually much better than MST. MST seems to be much more popular still for some reason I don't understand ...

xaviergonz:
oh well :-/

PR #435, opened 2022-03-17
add logo in SVG format

sisp:
I've created a logo in SVG format by manually re-drawing the PNG one. It isn't pixel-identical:
- The edges weren't perfectly sharp in the PNG version, so I had to guess the exact coordinates of the path points. The brackets are identical, the right one is the flipped version of the left one. The triangle in the middle is horizontally symmetric, the gaps are exactly parallel to the triangle's line on the left and have identical offsets to their neighbors. 
- The triangle in the middle wasn't perfectly centered vertically and possibly horizontally, but I assumed it should be, so I centered it.
- I tried to guess the parameters of the gradient by extracting its start and stop color and its angle, but it's not 100% identical to the PNG version.
- I cropped the canvas so the shape touches the boundaries of the viewport.

I've also updated the PNG version and the favicon, so they are all based on the SVG image.

What do you think?

Related to #433.


xaviergonz:
looks awesome!

PR #439, opened 2022-03-18
docs(patches): fix typo about the number of changes detection ways

### LINKED ISSUE #438: [Doc] Possible ways to detect changes ?
In the [Patches overview documentation](https://mobx-keystone.js.org/patches#overview), we can read:

> As seen in the previous snapshots section, any change made to a tree node will generate a new snapshot, but this is only one of the three possible ways mobx-keystone offers to detect changes. The second way is "patches".

So the first two ways to detect changes are snapshots and patches. Could you please indicate what is the third way?

gbcreation:
Here is a small PR to fix #438

### DIFF
--- a/packages/site/docs/patches.mdx
+++ b/packages/site/docs/patches.mdx
@@ -5,7 +5,7 @@ slug: /patches
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


xaviergonz:
Thanks!

Issue #452, opened 2022-05-08
React Native fails to resolve module due to mjs extension


kuasha420:
Hi, 

Great work with the library! Really liking it as a mst user. 

However I think I have discovered a packaging bug with the library when used in a react native project. The metro bundler can not resolve this module and throws the following error:

```
Error: While trying to resolve module `mobx-keystone` from file `/home/kuasha/Dev/RNKeystoneBug/src/stores/keystone.tsx`, the package `/home/kuasha/Dev/RNKeystoneBug/node_modules/mobx-keystone/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/home/kuasha/Dev/RNKeystoneBug/node_modules/mobx-keystone/dist/mobx-keystone.esm.mjs`. Indeed, none of these files exist:

  * /home/kuasha/Dev/RNKeystoneBug/node_modules/mobx-keystone/dist/mobx-keystone.esm.mjs(.native|.android.js|.native.js|.js|.android.json|.native.json|.json|.android.ts|.native.ts|.ts|.android.tsx|.native.tsx|.tsx)
  * /home/kuasha/Dev/RNKeystoneBug/node_modules/mobx-keystone/dist/mobx-keystone.esm.mjs/index(.native|.android.js|.native.js|.js|.android.json|.native.json|.json|.android.ts|.native.ts|.ts|.android.tsx|.native.tsx|.tsx)
``` 

This is because Metro Bundler does not recognize `.mjs` as a valid module file extension and tries to `creatively` resolve it. And in the `react-native` field of `package.json`, it points to the `.mjs` file. 

https://github.com/xaviergonz/mobx-keystone/blob/ecd6cbad9fcd3d9840f3a9ee63b48c68258cc25f/packages/lib/package.json#L37

I can confirm that all of the following method fixes the issue-

* Use the umd module with js extension. 
* Directly load the index.ts file from source 
* Rename the file from `mjs` to `js` in both filesystem and the `react-native` field of `package.json`.

ie. Either of the following works:
```
"react-native": "./src/index.ts",
// or
"react-native": "./dist/mobx-keystone.umd.js",
```

I think react native ecosystem prefers putting the actual un-transpiled source in the `react-native` field and it's done on the libs generated with callstacks' [react-native-builder-bob](https://github.com/callstack/react-native-builder-bob). However, I'll leave that decision to you. 

Would love to have it fixed soon so I can get rid of the patch-package solution. 

Cheers. 

[...]

PR #453, opened 2022-05-08
Fix react native module bug.

smamun19:
Addresses issue #452


xaviergonz:
Thanks! in the end I went for using the umd version (it feels kind of weird using a .ts file as entry point for a bundle...).
It should be fixed now in v0.68.4

smamun19:
SGTM. Cheers! ;)

Issue #460, opened 2022-05-28
Associating metadata with Model properties for runtime reflection?


kcoop:
Given a property name for a given model, I'm needing a mechanism to extract custom defined information about its associated property. I want, for example, to be able to associate custom formatter functions with different props, or to be able to mark a prop as required.

I could build a look aside table, but mobx-keystone's runtime types seem well suited to this, and it'd be nice to have everything about a prop be defined in a single place. I'm wondering if a) there's already existing code for looking up the type of a prop, and b), whether there's a way to add a generic data value to any runtime prop?

kcoop:
Ah, think I see how this could work, using composition, i.e. create a new `TypeInfo` and have it wrap an existing type, like `types.refinement` does.

And it looks like the mechanism in [typeCheck() ](https://github.com/xaviergonz/mobx-keystone/blob/f6ccf252190c4d2fb63e2b9ce1e68aa3515cc3c9/packages/lib/src/model/BaseModel.ts#L114) could be used to get the typeInfo:
```typescript
    const modelType = typesModel<this>(this.constructor as any)
   ...
```
but `typesModel()` isn't exported, so no joy. Seems like adding something like `Model.getTypeInfo(propKey: string)` would be a useful public function. Happy to create a pull request if you think this is the right approach.

kcoop:
Aha, looks like I can do this:

```typescript
@model('MyModel')
class MyModel extends Model({
  aPropValue: tProp(types.string, '')
}) {}

const type = types.model(MyModel);
const modelTypeInfo = getTypeInfo(type) as ModelTypeInfo;
const propTypeInfo = modelTypeInfo.props.aPropValue.typeInfo;
```

xaviergonz:
Yes, that's the right way to get the type info out of a model.

Why do you wanna avoid your own system to get custom data?

```
const customData = new WeakMap();

@model('MyModel')
class MyModel extends Model({
  aPropValue: tProp(types.string, '')
}) {}

customData.set(MyModel, {
  aPropValue: "beautiful"
})

const m = new MyModel(...)

customData.get(m.constructor) // { aPropValue: ... }
```

you could even type a setCustomData/getCustomData so it only accepts valid prop Names and so they do the m.constructor part auto if needed

kcoop:
Yes, not a terrible alternative, it's just that there's something nice about declaring everything about a prop in a single place.  Same reason decorators are nice, more readable, fewer footguns. 

[...]

PR #461, opened 2022-05-29
Issue #460 - add type.annotation

kcoop:
Implements a proposal for adding a new type, called type.annotation, that can wrap a prop's type to provide runtime access to metadata about the prop. See discussion [here]( https://github.com/xaviergonz/mobx-keystone/issues/460).

### DIFF
--- a/apps/site/docs/runtimeTypeChecking.mdx
+++ b/apps/site/docs/runtimeTypeChecking.mdx
@@ -282,10 +282,9 @@ const sumModelType = types.refinement(types.model(Sum), (sum) => {
 })
 ```
 
-### `types.annotation<T>(baseType, annotation: T)`
+### `types.annotation<T>(baseType, annotation: T, typeName?: string)`
 
-Wraps a given type with annotation information. This allows you to associate arbitrary metadata with the
-type of a prop that you can then use at runtime against instances.
+Wraps a given type with annotation information. This allows you to associate arbitrary metadata with the type of a prop that you can then use at runtime against instances.
 
 ```ts
 const widthType = types.annotation(
@@ -293,11 +292,13 @@ const widthType = types.annotation(
   { displayName: "Width in inches", required: true },
   "dimension"
 )
+
 const heightType = types.annotation(
   types.number,
   { displayName: "Height in inches", required: true },
   "dimension"
 )
+
 @model("MyModel")
 class MyModel extends Model({
   width: tProp(widthType, 10),
--- a/apps/site/docs/runtimeTypeChecking.mdx
+++ b/apps/site/docs/runtimeTypeChecking.mdx
@@ -282,18 +282,18 @@ const sumModelType = types.refinement(types.model(Sum), (sum) => {
 })
 ```
 
-### `types.annotation<T>(baseType, annotation: T, typeName?: string)`
+### `types.tag<T>(baseType, tag: T, typeName?: string)`
 
-Wraps a given type with annotation information. This allows you to associate arbitrary metadata with the type of a prop that you can then use at runtime against instances.
+Wraps a given type with tag information. This allows you to associate arbitrary metadata with the type of a prop that you can then use at runtime against instances.
 
 ```ts
-const widthType = types.annotation(
+const widthType = types.tag(
   types.number,
   { displayName: "Width in inches", required: true },
   "dimension"
 )
 
-const heightType = types.annotation(
+const heightType = types.tag(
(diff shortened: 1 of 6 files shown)


xaviergonz:
I have a question about the feature.

Do you think an annotation should be really a part of a type (`types.annotation(type, ...)`) or part of a property declaration (`prop/tProp(..., { annotation: {} })`)? Why do you prefer one over the other?

kcoop:
I actually would prefer it as a part of the property declaration, as that would be less verbose, and maybe more discoverable. 

I went that route because I thought it would be more palatable as an incremental change, since it's just the addition of an optional type rather than adding a new prop to the existing object model. 

If you're open to that approach, I'll make the change.

xaviergonz:
I am undefined because of this:
part of prop
  - prop pro: "seems" more logical to have it near the declared prop
  - prop pro/types cons: there are less corner cases (a prop either has an annotation or it doesn't, with types you could have a type with an annotation that has another annotation inside)
  - prop cons/types pro: no way to add annotation to anything else that is not a model prop (e.g. a plain object property, an array item, a model itself, etc.)

What's your real use case? (to try to figure out which of these weight more)

kcoop:
Initial use case is to generate React “field” components that have rich decoration (labels, required, etc.), and to be able to polymorphically describe values in written form.

Props will all be a part of models. Didn’t realize you could tProp a plain object, I’ve been working entirely with class based data.

kcoop:
Any further thoughts on this? Happy to put together a new pull request if you're open to part-of-prop approach (and it turns out not to be too disruptive implementation-wise).

kcoop:
One issue looking at the `prop()` function is the ambiguity of adding a second optional argument.

apps/site/docs/runtimeTypeChecking.mdx
```diff
@@ -282,6 +282,37 @@ const sumModelType = types.refinement(types.model(Sum), (sum) => {
 })
 ```
 
+### `types.annotation<T>(baseType, annotation: T)`
```
xaviergonz, line 285: the optional typename is missing

xaviergonz:
In the end I think types.annotation can just be used in more places and that out-weights props (even though that means it can only be used with tProps, but if you want reflection you probably want to use those).

Should I merge it?

kcoop:
Great idea, I like `withAnnotation()`. I'll look at this further later today, and update the PR.

xaviergonz:
Don't worry about updating the PR, it is OK as it is :) Like I said, I think using types is better since it allows you to deeply annotate any kind of type. E.g.:

```
const tPoint = types.object(() => ({
  x: types.annotation(types.number, "x coord"),
  y: types.annotation(types.number, "y coord"),
})

@model(...)
class M extends Model({
  point: tProp(types.annotation(tPoint, "a point")),
}) {
}
```

My only question was if you felt the PR was ready to be merged or if there was something else you wanted to add 👍

kcoop:
Yeah, I think it's ready to merge, can always add `.withAnnotation()` later. As it is it'll serve my needs well.

xaviergonz:
Actually, there's one thing now that I think about it. Would you consider types.tag a better name than types.annotation?

kcoop:
Or actually maybe 'data'?

kcoop:
Either one is fine by me.

kcoop:
Metadata's more descriptive, but long. Tag is fine. Do you want to make those changes, should I?

xaviergonz:
tag it is then. yes please, as soon as they are done I'll merge

sisp:
I just checked the docs for this new feature and couldn't find it there (https://mobx-keystone.js.org/runtime-type-checking/) although documentation was added in this PR. @xaviergonz Can you confirm the website hasn't been updated?

xaviergonz:
Thanks for the heads up, it wasn't due to the nodejs version being outdated. It should be ok in a few mins.

sisp:
Works now, thanks! :slightly_smiling_face:

kcoop:
Coming back to say I no longer think this is a good approach for the scenarios I originally envisioned. It involved tightly coupling model declaration with view related behavior, which is poor practice.

Since mobx-keystone provides unique type naming, a better approach is to use prop type discovery to acquire a type id, and then use that to map any view related attributes elsewhere.

I imagine there are still model-specific use cases for prop tags. If anyone has any suggestions for those, I'm happy to update the doc example to reflect a more appropriate usage.

PR #477, opened 2022-07-29
Fix: tag type wrapper check func captures wrong 'this'

kcoop:
Version 0.69.5 uncovered a bug in the `typesTag`: the baseChecker's check function was assigned directly rather than being wrapped within another function, which caused an endlessly recursive call to the check function from within itself because it captures `this` (i.e. `this.check()` invokes `_check()`, which is `this.check()`, which invokes `_check()`...).


xaviergonz:
Thanks!

PR #484, opened 2022-10-07
fix computed tree with array of models

### LINKED ISSUE #483: Array of models in computedTree
Does computedTree support an array of models?
I get an error "an object cannot be assigned a new parent when it already has one" when the tree is updated for the second time.
https://codesandbox.io/s/mobx-keystone-computedtree-cscdc

sisp:
I think it was actually incorrect to untweak `oldValue` and not `entry.tweakedValue` (i.e. `oldTweakedVlaue`) because `oldValue` hasn't been tweaked. I also saw this when stepping through the code with the debugger:

https://github.com/xaviergonz/mobx-keystone/blob/e35a792789fc66734b43cbf801d93a4e29ef9bd1/packages/lib/src/computedTree/computedTree.ts#L78

didn't execute

https://github.com/xaviergonz/mobx-keystone/blob/e35a792789fc66734b43cbf801d93a4e29ef9bd1/packages/lib/src/tweaker/tweak.ts#L91-L100

WDYT, @xaviergonz?

Fixes #483. :crossed_fingers:

### DIFF
--- a/packages/lib/src/computedTree/computedTree.ts
+++ b/packages/lib/src/computedTree/computedTree.ts
@@ -78,7 +78,6 @@ export function computedTree(
     const oldTweakedValue = entry.tweakedValue
     tweak(oldTweakedValue, undefined)
     tryUntweak(oldTweakedValue)
-    tryUntweak(oldValue)
 
     const tweakedValue = tweakComputedTreeNode(newValue, this, propertyKey)
     entry.value = newValue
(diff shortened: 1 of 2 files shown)


xaviergonz:
just wondering, do we need to keep tryUntweak(oldValue)?

sisp:
Argh, I forgot to remove this line in the end. You're right, that's not needed anymore.

xaviergonz:
Thanks a lot!

PR #548, opened 2024-09-12
fix: rename `runForCurrentChildren` to `fireForCurrentChildren`

yuqianma:
`fireForCurrentChildren: false` has no effect.

I should add corresponding tests but I'm busy at the moment. ( Sorry, but the bug is really clear. The fix should definitely work 😁. )


xaviergonz:
Thanks for the fix! I'll release a bugfix release right away

